PR #1, opened 2017-05-04
RUN-2863 [6.49.21.*] Start merging external process and external application APIs

nicholasdgoodman:
**To hold merge until 6.49.21**

Adapter Side PR: https://github.com/openfin/javascript-adapter/pull/298

New integration tests:
https://testing-dashboard.openfin.co/#/app/tests/590399b05114ac10ca6db30e/edit
https://testing-dashboard.openfin.co/#/app/tests/590398295114ac10ca6db30d/edit
https://testing-dashboard.openfin.co/#/app/tests/5903970d5114ac10ca6db30c/edit

Regression / integration test results:
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59038a515114ac10ca6db2fe
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59038e825114ac10ca6db2ff
(Additional runs)
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/590391445114ac10ca6db309

*FIRST!*


src/browser/api/external_application.ts
```diff
@@ -0,0 +1,85 @@
...
+            ofEvents.removeListener(evt, listener);
+        };
+    }
+
+    export function removeEventListener(identity: Identity, type: string, listener: Function) {
+        ofEvents.removeListener(`external-application/${type}/${identity.uuid}`, listener);
+    }
+
+    export function getInfo(externalApp: Identity): ExternalProcessInfo {
+        const process: any = ProcessTracker.getProcessByUuid(externalApp.uuid);
```
datamadic: `process` is already a [node global](https://nodejs.org/api/process.html), perhaps let's not shadow that?

src/browser/api/external_application.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    export function addExternalConnection(externalConnObj: Identity) {
+        const {
+            uuid
+        } = externalConnObj;
+
+        //TODO: compare perf from this and a map.
+        authenticatedConnections.push(externalConnObj);
+        ofEvents.emit(connectedEvent + `/${externalConnObj.uuid}`, {
```
datamadic: :neckbeard: just use the template literal here

src/browser/api/external_application.ts
```diff
@@ -0,0 +1,85 @@
...
+    export function getExternalConnectionById(id: number) {
+        return authenticatedConnections.find(c => {
+            return c.id === id;
+        });
+    }
+
+    export function removeExternalConnection(externalConnection: Identity) {
+        authenticatedConnections.splice(authenticatedConnections.indexOf(externalConnection), 1);
+
+        ofEvents.emit(disconnectedEvent + `/${externalConnection.uuid}`, {
```
datamadic: :neckbeard: just use the template literal here

src/browser/api_protocol/api_handlers/authorization.js
```diff
@@ -44,13 +47,31 @@ function AuthorizationApiHandler() {
...
-        var payload = message.payload,
-            uuid = payload.uuid,
-            file, token;
+
+        let {
+            uuid: uuidRequested,
+            pid
+        } = message.payload;
+
+        let process, file, token;
```
datamadic: :neckbeard: process is a node global

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -35,15 +35,21 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
+                // singleFrameOnly check first so to prevent frame superceding when disabled.
+                if (!data.singleFrameOnly === false || e.sender.isValidWithFrameConnect(e.frameRoutingId)) {
+                    Promise.resolve()
+                        .then(() => action(identity, data, ack, nack))
+                        .then(result => {
+                            if (result !== undefined) {
+                                ack(new AckPayload(result));
+                            }
+                        }).catch(err => {
+                            if (err !== undefined) {
```
datamadic: Shouldn't we be `nack`ing even if the `err` is undefined to guard against an empty `reject` call? 

By leaving it up to the `resolve` and `reject` payloads being defined (while I agree that they should be), by not calling the `ack` / `nack` you can potentially leave a renderer hanging in the worst case (a sync call) or at best leave a caller waiting forever (inviting a memory leak)

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -35,15 +35,21 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
-                        nack('API access has been superseded by another frame in this window.');
-                    }
-                } catch (err) {
-                    nack(err);
+                // singleFrameOnly check first so to prevent frame superceding when disabled.
+                if (!data.singleFrameOnly === false || e.sender.isValidWithFrameConnect(e.frameRoutingId)) {
+                    Promise.resolve()
+                        .then(() => action(identity, data, ack, nack))
+                        .then(result => {
+                            if (result !== undefined) {
```
datamadic: Shouldn't we be `ack`ing even if the result is `undefined` to guard against an empty `resolve` call?

src/browser/api_protocol/transport_strategy/ws_strategy.ts
```diff
@@ -27,12 +27,17 @@ export class WebSocketStrategy extends ApiTransportBase<MessagePackage> {
...
             } else if (typeof (action) === 'function') {
-
-                try {
-                    action(identity, data, ack, nack);
-                } catch (err) {
-                    nack(err);
-                }
+                Promise.resolve()
+                    .then(() => action(identity, data, ack, nack))
+                    .then(result => {
```
datamadic: See comments in `elipc_strategy.ts` on only ack / nack-ing if defined

src/browser/process_tracker.js
```diff
@@ -238,36 +260,57 @@ ProcessTracker.prototype.launch = function(winIdentity, config, resolve) {
...
+        uuid: uuidRequested,
+        lifetime,
+        monitor: monitorRequested,
+    } = options;
+
+    let pid = parseInt(pidRequested, 10);
+    let processEntry = this._processes[pid] || {};
+
+    if (isNaN(pid)) {
+        throw new Error(`Error monitoring external process, invalid pid value specified.`);
```
datamadic: With the `elipc_strategy.ts` this is sure to be caught right? Throwing on a call not initiated via an api call is not guaranteed to be handled gracefully.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
nicholasdgoodman: No. Almost all API handlers (except for the new external application ones) do not return a value. Rather they internally bear the responsibility of Ack'ing. To do so here would cause the vast majority of functions to Ack twice.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
nicholasdgoodman: As before. Not Ack'ing or Nack'ing when the functions do not return a value is preserving the status-quo. Previously it has been up to the implementors of each API call to ensure that one or the other has been called. The new paradigm of returning a value (or promise) lessens the likelihood of a dangling API call.

src/browser/process_tracker.js
nicholasdgoodman: Agree that this _may_ not be best practice overall, but throwing on argument validation was the existing coding style for this particular file. We can discuss this one.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
rdepena: in the unlikely case of a null err we still need to nack something, just create a new error and return it, silent fails are never good.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
rdepena: if we catch something we need to return something, if the error is undefined we still know that the action you requested failed. lets make sure we cover that.

PR #2, opened 2017-05-04
RUN-2807 reload breaks parenting model

datamadic:
For `executeJavaScript` you have to be in the parenting lineage of an app (or window). Restarting was breaking this because we were storing the parent info on the app obj that was being cleaned up on the close event. This PR moves the `parentUuid` up to the meta `app` object that persists through the close / restart.

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/590b263d5114ac10ca6db32e)
[added test](https://testing-dashboard.openfin.co/#/app/tests/58b5e799b4be9d42a7399632/edit)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -99,17 +99,17 @@ electronApp.on('ready', function() {
                 if (uuid) {
                     ofEvents.emit(eventRoute(uuid, 'manifest-changed'), sourceUrl, json);
                 } else {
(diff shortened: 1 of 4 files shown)


datamadic:
@HarsimranSingh your comments should be addressed with that last update

HarsimranSingh:
👍

rdepena:
👍

PR #8, opened 2017-05-08
RUN-2962 adds ability to subscribe to future cross-runtime events

whyn07m3:
This PR sets the base layer for supporting subscriptions to cross-runtime events for runtimes or apps that haven't been launched yet.

My next PR (programmatically launching apps via RVM) will depend on these code changes.

In the future it needs to be improved, but for now, it is known to work for needed events.

✅  [Tests OK](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5910c0105114ac10ca6db36e) (one is an inconsistent timeout, but doesn't come up when singled out)

### DIFF
--- a/src/browser/pending_subscriptions.ts
+++ b/src/browser/pending_subscriptions.ts
@@ -27,9 +27,9 @@ interface IPendingSubscriptionProps {
  * Shape of pending subscriptions
  */
 interface IPendingSubscription extends IPendingSubscriptionProps {
-    _id?: number; // ID of the subscription
-    isCleaned?: boolean; // helps prevents repetitive un-subscriptions
-    unSubscriptions?: { // a map of un-subscriptions assigned to runtime versions
+    _id: number; // ID of the subscription
+    isCleaned: boolean; // helps prevents repetitive un-subscriptions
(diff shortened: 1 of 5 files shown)


src/browser/pending_subscriptions.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/**
+ * Get event emitter of the class
+ */
+function getClassEventEmitter(runtime: PeerRuntime, subscription: IPendingSubscription) {
+    let classEventEmitter;
+    const { uuid, name, className } = subscription;
+
+    switch (className) {
```
HarsimranSingh: Needs a `default` case?

whyn07m3: Not needed, because it's an internal call and it's TypeScript ;) (above shape allows only certain class names `className: 'application'|'window'`, etc..), so `default` should/will never happen

rdepena:
👍

HarsimranSingh:
👍

whyn07m3:
✅  [Tests look good](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5916003d5114ac10ca6db3a7)

Good to go! 🚀

PR #12, opened 2017-05-09
RUN-2952 getDeviceUserId

datamadic:
Added `fin.desktop.System.getDeviceUserId` as a hash of the mac address and username. Works in both windows and mac (will not work for linux). 

Pairs with [this adapter PR](https://github.com/openfin/javascript-adapter/pull/306)

tests well (failing tests pass individually except [one preload test](https://testing-dashboard.openfin.co/#/app/tests/58e7d1a6067e1f3df99dd2e2/edit)) but I don't trust that that is this PR...
[added test](https://testing-dashboard.openfin.co/#/app/tests/591221aa5114ac10ca6db37a/edit)


wenjunche:
if (!username || !username) {

should be

 if (!username || !hostToken) {

datamadic:
@wenjunche yes, thanks, fixed

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -60,11 +60,19 @@ function SystemApiHandler() {
         'raise-event': raiseEvent,
         'download-asset': downloadAsset,
         'get-all-external-applications': getAllExternalApplications,
-        'resolve-uuid': resolveUuid
+        'resolve-uuid': resolveUuid,
+        'get-device-user-id': getDeviceUserId
```
HarsimranSingh: :neckbeard: put this in alphabetical order

src/browser/api/system.js
```diff
@@ -220,25 +221,50 @@ module.exports.System = {
...
+            username = process.env.USER;
+        } else {
+
+            // assume windows
+            hostToken = electronApp.getHostToken();
+            username = process.env.USERNAME;
+        }
+
+        if (!username || !hostToken) {
+            throw new Error(`One of username (${username}) or host token (${hostToken}) not defined `);
```
HarsimranSingh: Favor calling an error callback over throwing

HarsimranSingh:
Does this fail gracefully in linux?

src/browser/api/system.js
HarsimranSingh: :neckbeard: :neckbeard: extra space at the end of the message

src/browser/api_protocol/api_handlers/system.js
datamadic: will do

src/browser/api/system.js
datamadic: space yes, but I actually like throwing here, this is an API call where throwing will trigger the error callback for the caller

datamadic:
@HarsimranSingh there are a few linux options ([here](https://nodejs.org/api/process.html#process_process_platform)), I think for now its OK not to handle it. @StevenEBarbaro what do you think?

datamadic:
merged develop, [updated tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/591b53ab5114ac10ca6db3cc)

PR #14, opened 2017-05-11
[develop] backgroundThrottling should be false by default

HarsimranSingh:
@StevenEBarbaro

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -49,6 +49,7 @@ function five0BaseOptions() {
         'alwaysOnTop': false,
         'applicationIcon': '',
         'autoShow': false,
+        'backgroundThrottling': false,
         'contextMenu': true,
         'cornerRounding': {
             'height': 0,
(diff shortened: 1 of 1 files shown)


src/browser/convert_options.js
```diff
@@ -266,10 +267,6 @@ module.exports = {
             newOptions.preload = options.preload;
         }
 
-        if (options.hasOwnProperty('backgroundThrottling')) {
-            newOptions.backgroundThrottling = options.backgroundThrottling;
-        }
```
whyn07m3: I believe this needs to be kept here, in addition to the added default

HarsimranSingh: If it's a part of the base options then this code isn't needed. Adding window options like this actually should be discouraged and only used for options that are used internally, like notifications passing around the notification text [here](https://github.com/HadoukenIO/core/blob/develop/src/browser/convert_options.js#L253).

rdepena:
👍

HarsimranSingh:
Will update with tests

HarsimranSingh:
[Clean Test Run](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/591c58ff5114ac10ca6db3dd)

src/browser/convert_options.js
whyn07m3: @HarsimranSingh Got it, didn't know it was not an option, just the default

PR #19, opened 2017-05-15
RUN-2098 add better support for fractional bounds

rdepena:
@wenjunche @StevenEBarbaro @HarsimranSingh @datamadic  @joneit @whyn07m3 

- [x] Fixed bound calls (moveBy, moveTo, resizeBy, resizeTo, setBounds, showAt) to handle Fractional bounds
- [x] Fixed grunt test task to run everything up to the ts build before testing.
- [x] [New Unit tests](https://github.com/HadoukenIO/core/compare/develop...rdepena:feature/RUN-2098_fractional_bounds?expand=1#diff-9fa45cc1a31a84eae9e9caa77f50fe58R22)
- [x] [Test results](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/591a1e2e5114ac10ca6db3c3)


src/common/safe_int.ts
```diff
@@ -0,0 +1,29 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export function toSafeInt(n: number, def?: number) {
+    if (Number.isSafeInteger(n)) {
```
datamadic: :neckbeard: Should we also add a simple `typeof` check here as well? Even though this is TS, it is getting called from JS at the moment. You are not assured that you are getting a `number`. For example, this will accept `boolean` and empty arrays as the `n` value without triggering the default value or throwing.

rdepena: thats a good point, isNaN is pretty loose, I think I will just switch it to typeof, test-rerun pending.

rdepena:
[Added new unit tests for non numbers](https://github.com/HadoukenIO/core/pull/19/files#diff-9fa45cc1a31a84eae9e9caa77f50fe58R37)
[Updated Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/591ef96d5114ac10ca6db433)

PR #25, opened 2017-05-19
Updated the runtime-p2p dependency that includes the auth issue fix on install.

rdepena:
@whyn07m3 Fixed a dependency issue in the [runtime-p2p repo](https://github.com/openfin/runtime-p2p/commit/91ed159a9e49be32fb794f4be3ef4a9e9f0f01b9), bringing that into the fold.


HarsimranSingh:
👍

HarsimranSingh:
Question: why doesn't `runtime-p2p` also have a `git+ssh://` prefix like `openfin-sign`?

rdepena:
@HarsimranSingh not really sure. 

[updated tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59270bc65114ac10ca6db47d)

HarsimranSingh:
https://docs.npmjs.com/cli/install

This leads me to believe that if you don't specify a protocol (like `runtime-p2p`) then it defaults to `https://`.

It doesn't explain why you need a `git+` before it though...

PR #29, opened 2017-05-23
RUN-3017 normalize subscriber event constants

joneit:
[Test](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5924e4af5114ac10ca6db456) was clean — except for _Notification.close (from onShow)_ failing + two fails which passed on retest.

### DIFF
--- a/src/browser/of_events.ts
+++ b/src/browser/of_events.ts
@@ -43,7 +43,7 @@ class OFEvents extends EventEmitter {
         return super.emit(route, ...data);
     }
 
-    public SUBSCRIBER: StringMap = {
+    public subscriber: StringMap = {
         ADDED: 'subscriber-added',
         REMOVED: 'subscriber-removed'
     };
--- a/src/browser/of_events.ts
+++ b/src/browser/of_events.ts
@@ -42,6 +42,15 @@ class OFEvents extends EventEmitter {
 
         return super.emit(route, ...data);
     }
+
+    public SUBSCRIBER: StringMap = {
+        ADDED: 'subscriber-added',
+        REMOVED: 'subscriber-removed'
+    };
+}
+
+interface StringMap {
+    [key: string]: string;
 }
 
 export default new OFEvents();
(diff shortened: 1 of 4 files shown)


HarsimranSingh:
👍

src/browser/of_events.ts
```diff
@@ -42,6 +42,15 @@ class OFEvents extends EventEmitter {
 
         return super.emit(route, ...data);
     }
+
+    public SUBSCRIBER: StringMap = {
```
datamadic: ❓  Does this make sense as a static property?

whyn07m3: Same here, static? Also, why uppercase, I mean, I kinda get it why, but then again... "why"?

joneit: Regarding static, note the last line:
```js
export default new OFEvents();
```
So the 14 files that import this only have the instantiated object handy and not the constructor. Instead of `ofEvents.SUBSCRIBER.ADDED` we would have to have `ofEvents.constructor.SUBSCRIBER.ADDED`. This does not IMHO to enhance readability.

joneit: Regarding UPPERCASE, yea it's constants. Too old school?
```js
ofEvents.SUBSCRIBER.ADDED
ofEvents.SUBSCRIBER.added
ofEvents.subscriber.ADDED
ofEvents.subscriber.added
```
I like 1; I could live with 3; I don't like 2 and 4.

datamadic: So for `static`, you would also `import` the `OfEvents` constructor itself which would have the static members already so
````js
import {default as whatever, OfEvents} from '../of_events';

ofEvents.SUBSCRIBER.ADDED // ... 
````

But you have convinced me that its fine as is 😄

joneit: yea trying not to edit 14 files just to introduce `OfEvents` which (due to o vs O) looks so much like `ofEvents`...

Regarding casing, I'll change it to option 3.

PR #32, opened 2017-05-24
RUN-2966 webSecurity hookup and cleanup

HarsimranSingh:
So `webSecurity` was already hooked up, but it wasn't documented. This is more than anything a cleanup of   `convert_options.js` and `index.js`.

Adapter PR: https://github.com/openfin/javascript-adapter/pull/311

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -13,23 +13,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-/*
-    src/browser/convert_options.js
- */
 
 // built-in modules
-let fs = require('fs');
-let path = require('path');
+const fs = require('fs');
+const path = require('path');
 
-let ResourceFetcher = require('electron').resourceFetcher;
+const ResourceFetcher = require('electron').resourceFetcher;
 
 // npm modules
-let _ = require('underscore');
+const _ = require('underscore');
 
 // local modules
-let coreState = require('./core_state.js');
-let log = require('./log');
-let regex = require('../common/regex');
+const coreState = require('./core_state.js');
+import * as log from './log';
+import * as regex from '../common/regex';
 
 // this is the 5.0 base to be sure that we are only extending what is already expected
 function five0BaseOptions() {
@@ -49,12 +46,14 @@ function five0BaseOptions() {
         'alwaysOnTop': false,
         'applicationIcon': '',
         'autoShow': false,
+        'backgroundColor': '#000',
         'backgroundThrottling': false,
         'contextMenu': true,
         'cornerRounding': {
             'height': 0,
             'width': 0
         },
+        'customData': '',
         'defaultCentered': false,
         'defaultHeight': 500,
         'defaultLeft': 10,
@@ -79,7 +78,9 @@ function five0BaseOptions() {
         'minimizable': true,
         'name': '',
         'opacity': 1,
+        'permissions': null,
         'plugins': false,
+        'preload': '',
         'resizable': true,
         'resize': true,
         'resizeRegion': {
@@ -96,12 +97,8 @@ function five0BaseOptions() {
         'url': 'about:blank',
         'uuid': '',
         'waitForPageLoad': true,
-        'backgroundColor': '#000'
+        'webSecurity': true
     };
-}
-
-function isInContainer(type) {
-    return process && process.versions && process.versions[type];
 }
 
 function readFile(filePath, done, onError) {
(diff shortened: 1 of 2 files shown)


wenjunche:
newOptions.webSecurity && !coreState.argo['disable-web-security']

should we allow desktop owners to enforce this ?

src/browser/convert_options.js
```diff
@@ -13,23 +13,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-let _ = require('underscore');
+const _ = require('underscore');
 
 // local modules
-let coreState = require('./core_state.js');
-let log = require('./log');
-let regex = require('../common/regex');
+const coreState = require('./core_state.js');
+import * as log from './log';
+import * as regex from '../common/regex';
```
whyn07m3: Can we also `import` (instead of `require`) other modules as well?

src/browser/convert_options.js
```diff
@@ -195,87 +180,66 @@ function fetchLocalConfig(configUrl, successCallback, errorCallback) {
...
+        const newOptions = validate(five0BaseOptions(), options);
+
+        newOptions.resizable = newOptions.resize && newOptions.resizable;
+        newOptions.show = newOptions.autoShow && !newOptions.waitForPageLoad;
+        newOptions.skipTaskbar = !newOptions.showTaskbarIcon;
+        newOptions.title = newOptions.name;
+
+        const minHeight = newOptions.minHeight;
+        const maxHeight = newOptions.maxHeight;
+        const defaultHeight = newOptions.defaultHeight;
```
whyn07m3: Can we move these to the top and de-structure them?

src/browser/convert_options.js
```diff
@@ -195,87 +180,66 @@ function fetchLocalConfig(configUrl, successCallback, errorCallback) {
...
-            if (defaultWidth < minWidth) {
-                newOptions.width = minWidth;
-            } else if (maxWidth !== -1 && defaultWidth > maxWidth) {
-                newOptions.width = maxWidth;
-            } else {
-                newOptions.width = defaultWidth;
-            }
+        const defaultWidth = newOptions.defaultWidth;
+        const minWidth = newOptions.minWidth;
+        const maxWidth = newOptions.maxWidth;
```
whyn07m3: Can we move these to the top and de-structure them?

src/browser/convert_options.js
```diff
@@ -195,87 +180,66 @@ function fetchLocalConfig(configUrl, successCallback, errorCallback) {
...
-            newOptions.customData = options.customData;
-        }
-
-        if (options.permissions !== undefined) { // API policy
-            newOptions.permissions = options.permissions;
-        }
-
-        if (options.hasOwnProperty('preload')) {
-            newOptions.preload = options.preload;
-        }
```
whyn07m3: Since this is being removed, do we check `hasOwnProperty` anywhere else in the code, versus checking for truthy values?

joneit: to be really safe and avoid the question entirely, we might make a little more use of prototype-less objects:
```js
var obj = Obect.create(null, {...})
```
In this case, add it around the return value in `five0BaseOptions`:
```js
function five0BaseOptions() {
    return Object.create(null, {
        ...
    });
}
```

HarsimranSingh: 👍

src/browser/convert_options.js
whyn07m3: No? :)

src/browser/convert_options.js
whyn07m3: No? :)

HarsimranSingh: 👍

src/browser/convert_options.js
HarsimranSingh: 👍

src/browser/convert_options.js
HarsimranSingh: Are you suggesting all the modules be imported or just `coreState`? It seems like currently we're only importing modules that have made the TS transition.

whyn07m3: I see, my bad

src/browser/convert_options.js
HarsimranSingh: So we don't just check for truthiness, we actually compare that the types are the same https://github.com/HarsimranSingh/core/blob/d652f83960c39f87443350f10cb335700e6e3b0f/src/browser/convert_options.js#L152

HarsimranSingh:
Updated with changes requested by @whyn07m3 and @joneit

src/browser/convert_options.js
whyn07m3: I don't fully trust `validate` function, as it won't update options with different types. Not sure we use different types for any of the options now, but maybe we will in the future and this will become a problem.
Also, arrays become problems too, as they are objects as well.

HarsimranSingh:
Closing since the refactor could use some more thought. Also this should probably be converted to typescript

PR #35, opened 2017-05-25
RUN-3013: Resizeable gets lost on maximize + restore

joneit:
Tested manually.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1295,6 +1295,12 @@ Window.restore = function(identity) {
         browserWindow.restore();
     } else if (browserWindow.isMaximized()) {
         browserWindow.unmaximize();
+        let resizable = getOptFromBrowserWin('resizable', browserWindow);
+        if (resizable !== undefined) {
+            resizable = !!resizable;
+            browserWindow.setResizable(resizable);
+            setOptOnBrowserWin('resizable', resizable, browserWindow);
+        }
     } else {
         browserWindow.showInactive();
     }


HarsimranSingh:
Not sure this accounts for double clicking the top of the window to restore.

Also, the 18 stable release behaves correctly, but something broke between then and now. That needs more investigation.

joneit:
Just noticed I apparently closed this. Didn't mean to!

joneit:
@HarsimranSingh Regarding double-click on title bar:
* Double-clicking to maximize does nothing, as expected when the window is supposed to be non-resizable.
* Double-clicking programmatically maximized window does respond by restoring window to pre-maximized size. **As you suspected, the edge arrows are now active.**

joneit:
@HarsimranSingh Regarding
> something broke between then and now
Hello Openfin using this fix rebased to tip of `develop` & deployed on top of .21.11 is working as expected.... what exactly "broke?"

joneit:
@HarsimranSingh Regarding double-click on title bar:
* Double-clicking to maximize does nothing, as expected when the window is supposed to be non-resizable.
* Double-clicking programmatically maximized window does respond by unmaximizing window (restoring to pre-maximized size) but **as you suspected,** our `restore` function apparently is not being called as the edge arrows are now active.

joneit:
Use cases to demaximize a maximized non-resizable window:
1. calling `fin.desktop.Window.getCurrent().restore()`
2. double-clicking in the title bar

Closing because:
* We determined that both demaximize use cases were previously working in .18.41 _without the additional code contained in this PR)
* We need a runtime-size fix anyway to deal with use case #2, so let's just figure out what changed on the runtime side and fixing that will probably fix both use cases without any additional javascript.

PR #41, opened 2017-05-31
RUN-2846 license info from app launch / adapters

datamadic:
Send the license key as well as some predefined meta info upon app launch and external connection authorization. 

This is ahead of the RVM's ability to accept such info. Talking to Omer and Pat, this should not be an issue as these new messages will should be dropped upon arrival. 

Also some very basic local cleanup (removal of the /* callback, errorCallback */ silliness). 

No new end-to-end tests, basic shape checks are done with simple unit tests / mocks. 

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/592f0d27998bc04e20cc6c94) 
Tests that fail are either pegged to the 7.53.21.* release (create from manifest...) or are failing even on fresh 7.53.20.20 for me (window navigation... )

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -329,7 +329,7 @@ Application.getShortcuts = function(identity, callback, errorCallback) {
         .catch(errorCallback);
 };
 
-Application.getInfo = function(identity, callback /*, errorCallback*/ ) {
+Application.getInfo = function(identity, callback) {
     const app = Application.wrap(identity.uuid);
 
     const response = {
@@ -345,17 +345,17 @@ Application.getWindow = function(identity) {
     return Window.wrap(uuid, uuid);
 };
 
-Application.grantAccess = function( /*action, callback, errorCallback*/ ) {
+Application.grantAccess = function() {
     console.warn('Deprecated');
 };
-Application.grantWindowAccess = function( /*action, windowName, callback, errorCallback*/ ) {
+Application.grantWindowAccess = function() {
     console.warn('Deprecated');
 };
-Application.isRunning = function(identity /*, callback, errorCallback*/ ) {
+Application.isRunning = function(identity) {
     let uuid = identity && identity.uuid;
     return uuid && coreState.getAppRunningState(uuid) && !coreState.getAppRestartingState(uuid);
 };
-Application.pingChildWindow = function( /*name, callback, errorCallback*/ ) {
+Application.pingChildWindow = function() {
     console.warn('Deprecated');
 };
 Application.registerCustomData = function(identity, data, callback, errorCallback) {
@@ -385,19 +385,19 @@ Application.registerCustomData = function(identity, data, callback, errorCallbac
 };
 
 //TODO:Ricardo: This should be deprecated.
-Application.removeEventListener = function(identity, type, listener /*, callback, errorCallback*/ ) {
+Application.removeEventListener = function(identity, type, listener) {
     var app = Application.wrap(identity.uuid);
 
     ofEvents.removeListener(eventRoute(app.id, type), listener);
 };
 
-Application.removeTrayIcon = function(identity /*, callback, errorCallback*/ ) {
+Application.removeTrayIcon = function(identity) {
     const app = Application.wrap(identity.uuid);
 
     removeTrayIcon(app);
 };
 
-Application.restart = function(identity /*, callback, errorCallback*/ ) {
+Application.restart = function(identity) {
     let uuid = identity.uuid;
     const appObj = coreState.getAppObjByUuid(uuid);
 
@@ -417,14 +417,17 @@ Application.restart = function(identity /*, callback, errorCallback*/ ) {
         throw err;
     }
 };
-Application.revokeAccess = function( /*action, callback, errorCallback*/ ) {
-    console.warn('Deprecated');
-};
-Application.revokeWindowAccess = function( /*action, windowName, callback, errorCallback*/ ) {
+
+Application.revokeAccess = function() {
     console.warn('Deprecated');
 };
 
-Application.run = function(identity, configUrl = '' /*, callback , errorCallback*/ ) {
+Application.revokeWindowAccess = function() {
+    console.warn('Deprecated');
+};
+
+Application.run = function(identity, configUrl = '') {
+
     if (!identity) {
         return;
     }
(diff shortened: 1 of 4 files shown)


src/browser/api/application.js
```diff
@@ -86,8 +86,9 @@ electronApp.on('use-plugins-requested', event => {
 });
 
 electronApp.on('ready', function() {
-    log.writeToLog(1, 'RVM MESSAGE BUS READY', true);
     rvmBus = require('../rvm/rvm_message_bus').rvmMessageBus;
+    log.writeToLog(1, 'RVM MESSAGE BUS READY', true);
```
whyn07m3: this is so good :)

src/browser/api/application.js
```diff
@@ -115,47 +116,53 @@ Application.create = function(opts, configUrl = '', parentIdentity = {}) {
...
     }
 
-    let appObj = createAppObj(opts.uuid, opts, configUrl);
+    const appObj = createAppObj(uuid, opts, configUrl);
 
     if (parentIdentity && parentIdentity.uuid) {
-        let app = coreState.appByUuid(opts.uuid);
 
-        app.parentUuid = parentIdentity.uuid;
+        // This a reference to the meta `app` object that is stored in core state,
```
whyn07m3: This **is** a reference... 😜

src/browser/api_protocol/api_handlers/authorization.js
```diff
@@ -15,7 +15,7 @@ let _ = require('underscore');
 let log = require('../../log');
 let socketServer = require('../../transports/socket_server').server;
 let ProcessTracker = require('../../process_tracker.js');
-
+const rvmMessageBus = require('../../rvm/rvm_message_bus').rvmMessageBus;
```
whyn07m3: Can we `import` it?

test/rvm_message_bus.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import {EventEmitter} from 'events';
+const App = require('electron').app;
+
+App.generateGUID = Math.random;
+
+const mockWMCopyData  = {
+    WMCopyData: function(){
+        return {
+            on: (x: any) => x,
+            publish: (x: any ) => x
```
whyn07m3: We need to add `/test` folder to our linting

test/rvm_message_bus.test.ts
```diff
@@ -0,0 +1,110 @@
...
+*/
+import * as assert from 'assert';
+import * as mockery from 'mockery';
+import {EventEmitter} from 'events';
+const App = require('electron').app;
+
+App.generateGUID = Math.random;
+
+const mockWMCopyData  = {
+    WMCopyData: function(){
```
whyn07m3: Can it be `WMCopyData: () => {`?

test/rvm_message_bus.test.ts
```diff
@@ -0,0 +1,110 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import * as assert from 'assert';
+import * as mockery from 'mockery';
+import {EventEmitter} from 'events';
+const App = require('electron').app;
```
whyn07m3: Maybe `import` here?

test/rvm_message_bus.test.ts
```diff
@@ -0,0 +1,110 @@
...
+            on: (x: any) => x,
+            publish: (x: any ) => x
+        }
+    }
+}
+
+mockery.registerMock('../transport', mockWMCopyData);
+mockery.enable();
+
+import {rvmMessageBus, RVMMessageBus, LicenseInfo} from '../src/browser/rvm/rvm_message_bus';
```
whyn07m3: Let's put it on top?

test/rvm_message_bus.test.ts
```diff
@@ -0,0 +1,110 @@
...
+                    version: null,
+                    pid: null
+                }
+            }
+            const {payload} = <any> rvmMessageBus.registerLicenseInfo(<LicenseInfo> {
+                processId: 'processId',
+                runtimeVersion: 'runtimeVersion'
+            });
+
+            assert.deepEqual(payloadShape, payload, 'shapes should match hommie');
```
whyn07m3: Let's keep it professional :bowtie:

src/browser/api/application.js
```diff
@@ -524,6 +535,21 @@ Application.run = function(identity, configUrl = '' /*, callback , errorCallback
             type: 'connected',
             uuid
         });
+
+        const parentConfigUrl = coreState.getConfigUrlByUuid(uuid);
+
+        rvmBus.registerLicenseInfo({
+            licenseKey: mainWindowOpts.licenseKey,
```
HarsimranSingh: `licenseKey` isn't at the options level, it's at the top app config level i think

src/browser/api/application.js
datamadic: it was there...

src/browser/api/application.js
datamadic: will fix

src/browser/api_protocol/api_handlers/authorization.js
datamadic: meh, this is still js, if we want to get off Babel one day pre TS conversion we will  just have to switch it back

test/rvm_message_bus.test.ts
datamadic: if we do then we have to deal with `generateGUID` not being in the d.ts file...

test/rvm_message_bus.test.ts
datamadic: yes, will fix

test/rvm_message_bus.test.ts
datamadic: There quite a bit of additional linting here... https://appoji.jira.com/browse/RUN-3035

test/rvm_message_bus.test.ts
datamadic: needs to be here for testing reasons

test/rvm_message_bus.test.ts
datamadic: yes, will fix

src/browser/api/application.js
datamadic: `mainWindowOpts` here is directly the `_options` object here (bad naming I guess, its what was there...) so this should be the full option set

src/browser/rvm/rvm_message_bus.ts
```diff
@@ -147,15 +147,28 @@ export interface ApplicationEvent extends RvmMsgBase {
     sourceUrl: string;
 }
 
+export interface LicenseInfo {
+    licenseKey?: string;
+    client?: {
+        type: 'dotnet' | 'java' | 'air' | 'node' | 'js';
```
datamadic: For `js` it will be internal. This is also servers as a point of truth or if / when validation is added

src/browser/api/application.js
datamadic: this should be the full option set here, `mainWindowOpts` was just a bad existing variable name

test/rvm_message_bus.test.ts
whyn07m3: We have a file where we add these kind of things (like `generateGUID`)
[Check it out](https://github.com/HadoukenIO/core/blob/develop/src/modules.d.ts#L22)

PR #37, opened 2017-05-26
RUN-3020: Normalize route string fabrication

joneit:
Yes, 21 files, but **Don't Panic!** because most of the edits one-to-one conversion from route literal to the new call.

I reviewed all the calls for accuracy. [Tested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59275fc5998bc04e20cc6c8b) clean.

**Caveat:** The only literal route strings remaining in the code are in `src/renderer/api-decorator.js` (because of that file has no access to `require()`).

### New file: `src/common/routes.ts`
Generic route:
```js
'channel/topic/subtopic/subsubtopic'
```
Notes:
* Sometimes in our code "channel" is called `topic`; and "topic" is called `type`
* Window routes often end in `uuid/name`; they very occasionally end in `uuid-name`; and sometimes they end in something entirely different

### Features
* calling syntax designed for convenience/readability
* although `route` not a verb (or at least not intended to be so), I like it because it's short & sweet
* `useHyphen` param changes delimiter between 3rd and 4th parts from default to `'-'`
* note that default for `useHyphen` is `false` in general but is `true` for window (and external window)

In addition to the generic `route` function, I've defined several convenience subfunctions:
```js
route(channel, topic, subtopic, subsubtopic, useHyphen=false)
route.application(topic, subtopic, subsubtopic); // channel defaults to 'application'
route.window(topic, uuid, name, useHyphen=true); // channel defaults to 'window`, etc.
route.externalWindow(topic, uuid, name, useHyphen=true);
// etc. for route.system, .rvmMessageBus, .server, .connection, .runtime
```

Use the generic `route` function for:
* Programmaticaly fabricated routes where `channel` is indeterminate until execution time
* When the channel is not accommodated by one of the predefined subfunctions
```js
var extConnRoute = route('externalconn', 'closed'); // -> 'externalconn/closed'
```

Use the generic `route` function for:
* Programmaticaly fabricated routes where `channel` is indeterminate until execution time
* When the channel is not accommodated by one of the predefined subfunctions
```js
var extConnRoute = route('externalconn', 'closed'); // -> 'externalconn/closed'
```

### Examples
```js
// generic route calls default to NOT use hyphen (include true to use)
route('window', 'close-requested', uuid, name); // `notifications/listener/${uuid}/${name}`
route('window', 'unload', uuid, name, true); // `notifications/listener/${uuid}-${name}`

// window-specific routes use final hyphen by default (include false to not use)
[...]

PR #74, opened 2017-06-23
RUN-3030: Abstract out some repetitive route and emit call patterns

joneit:
Added some local abstractions:
* `routeWin`
* `routeExtWin`
* `emitToApp`
* `emitToExtWin`

**Feedback please your comfort level with merging these now _vs._ living with new `route` functions a bit longer before we add these additional layers of abstraction.** We could easily postpone merge to a future sprint.

When inventing these, please keep to the above naming conventions, _i.e.,_ begin with either `route` or `emitTo`. This will make routes easier to search for globally using the regex I recommended in #37.

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -458,6 +458,7 @@ Application.run = function(identity, configUrl = '') {
     const mainWindowOpts = _.clone(app._options);
     const hideSplashTopic = route.application('hide-splashscreen', uuid);
     const eventListenerStrings = [];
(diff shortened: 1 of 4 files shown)


rdepena:
not sure we want this. I am comfortable with the current abstraction. 

👎

rdepena:
In fact, the first time you proposed these changes I had the same feedback:

https://github.com/HadoukenIO/core/pull/37/files/faecc30a678d440310d36ec065a0e59c7a792c2a#diff-d5752a5b2c3b7ce27c77d78f1b9278a6R414

PR #75, opened 2017-06-23
 RUN-2989 bad port discovery message / no error dialogs 

datamadic:
The port discovery message will now only be sent out if the initial app load is successful. A typo in the `noerrdialogs` guard was fixed as well. 

@wenjunche I tested with .NET and Node adapters and it seems to behave, can you check against Java? 
@rdepena You had expressed some concern with moving the message here, can you remind me what it was 😄? 

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/594d3a830c10170217365886) all fails pass in isolation

### DIFF
--- a/index.js
+++ b/index.js
@@ -313,6 +313,7 @@ app.on('ready', function() {
             });
         }
     });
+
     rvmBus.on(route.rvmMessageBus('broadcast', 'download-asset', 'complete'), payload => {
         if (payload) {
             ofEvents.emit(route.system(`asset-download-complete-${payload.downloadId}`), {
@@ -490,7 +491,7 @@ function launchApp(argo, startExternalAdapterServer) {
         if (openfinWinOpts && !isRunning) {
             //making sure that if a window is pressent we set the window name === to the uuid as per 5.0
             openfinWinOpts.name = uuid;
-            initFirstApp(openfinWinOpts, configUrl, licenseKey);
+            initFirstApp(configObject, configUrl, licenseKey);
         } else if (uuid) {
             Application.run({
                 uuid,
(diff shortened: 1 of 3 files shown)


HarsimranSingh:
👍

datamadic:
Conflicts resolved and retested
[updated test run](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59515c140c101702173658a7) parody with stable / fail passes in isolation

HarsimranSingh:
@datamadic can you rebase off develop and kick off a test run using 7.53.22.5

datamadic:
[updated/rebased tests 7.53.22.5](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5952c04b0c101702173658c7) fails pass in isolation 

@HarsimranSingh

index.js
```diff
@@ -514,7 +514,7 @@ function launchApp(argo, startExternalAdapterServer) {
         if (openfinWinOpts && !isRunning) {
             //making sure that if a window is pressent we set the window name === to the uuid as per 5.0
             openfinWinOpts.name = uuid;
```
nicholasdgoodman: This statement still succeeds only because of the nature of `getWindowOptions(...)` (terribly named function, I might add, if you look at what the function actually does).

It is equivalent to:

```
configObject.startup_app.name = uuid;
```

As written, the code is not exactly clear / maintainable, because it requires understanding the implementation details of `getWindowOptions(...)` function.

Also, it is brittle code: if the `getWindowOptions` call were ever refactored to return a copy of the object, for example, the assignment would fail.

**Suggestion**: If setting the name = to UUID is explicitly for 5.0 compatibility, doesn't that behavior belong in the `convertOptions` utility? Isn't that its entire purpose?

datamadic: 👍 @nicholasdgoodman have a look at the update

[updated tests look good](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5953b00b0c101702173658d6) fails pass in isolation

nicholasdgoodman: i like! _because how things are named is important_.

PR #91, opened 2017-07-17
RUN-3145 Emit diagnostic events from Window

joneit:
1. Moved code from `Application.run()` to `new Window()` (constructor) that handles certain Electron "diagnostic" events, re-emiting them to the app (as `'crashed'`, `'responding'`, and `'not-responding'`).
2. Doubled each emit to also emit to window, routed with `-`_name_ and with _name_ added to event object (where _name_ is the window name).

Testing this was a little challenging but I finally came up with a manual test. It is not practical to test this in the test runner because (1) per Roma we do not want to crash the test runner and (2) anyway my test involved manual typing directly into the debugger.

See Jira [RUN-3145](https://appoji.jira.com/browse/RUN-3145) for more details, including my test recipe.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -477,7 +477,7 @@ Window.create = function(id, opts) {
             });
 
             // can't unhook when the 'closed' event fires; browserWindow is already destroyed then
-            webContents.removeAllListeners('page-favicon-updated');
+            browserWindow.webContents.removeAllListeners('page-favicon-updated');
 
             // make sure that this uuid/name combo does not have any lingering close-requested subscriptions.
             ofEvents.removeAllListeners(closeEventString);
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -504,10 +504,17 @@ Window.create = function(id, opts) {
 
         const emitToAppAndWin = (...types) => {
             let isMainWindow = (uuid === name);
+
             types.forEach(type => {
+                // Window crashed: inform Window "namespace"
                 ofEvents.emit(route.window(type, uuid, name), { topic: 'window', type, uuid, name });
+
+                // Window crashed: inform Application "namespace" but with "window-" event string prefix
+                ofEvents.emit(route.application(`window-${type}`, uuid), { topic: 'application', type, uuid, name });
+
                 if (isMainWindow) {
-                    ofEvents.emit(route.application(`window-${type}`, uuid), { topic: 'application', type, uuid, name });
+                    // Application crashed: inform Application "namespace"
+                    ofEvents.emit(route.application(type, uuid), { topic: 'application', type, uuid });
                 }
             });
         };
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -502,10 +502,13 @@ Window.create = function(id, opts) {
             }
         });
 
-        var emitToAppAndWin = (...types) => {
+        const emitToAppAndWin = (...types) => {
+            let isMainWindow = (uuid === name);
             types.forEach(type => {
                 ofEvents.emit(route.window(type, uuid, name), { topic: 'window', type, uuid, name });
-                ofEvents.emit(route.application(`window-${type}`, uuid), { topic: 'application', type, uuid, name });
+                if (isMainWindow) {
+                    ofEvents.emit(route.application(`window-${type}`, uuid), { topic: 'application', type, uuid, name });
+                }
             });
         };
 
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -373,7 +373,7 @@ Window.create = function(id, opts) {
 
     let _externalWindowEventAdapter;
 
-    // we need to be able to handle the wrapped case, ie. dont try to
+    // we need to be able to handle the wrapped case, ie. don't try to
     // grab the browser window instance because it may not exist, or
     // perhaps just try ...
     if (!opts._noregister) {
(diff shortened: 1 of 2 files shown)


src/browser/api/window.js
```diff
@@ -502,6 +502,25 @@ Window.create = function(id, opts) {
             }
         });
 
+        var emitToAppAndWin = (...types) => {
+            types.forEach(type => {
+                ofEvents.emit(route.window(type, uuid, name), { topic: 'window', type, uuid, name });
+                ofEvents.emit(route.application(type, uuid), { topic: 'application', type, uuid });
```
rdepena: Lets add the window name here, in an app that has multi origin windows we will need the window name to be able to determine what crashed.

katierose4:
I spoke about this with @StevenEBarbaro, but this change needs to make it in the **v6 hotfix** as this is an upgrade blocker for Barclays. /cc @nicholasdgoodman @rdepena

rdepena:
@katierose4 cool, once these changes make their way into v7 we can cherry pick and PR into 6/Staging.

src/browser/api/window.js
```diff
@@ -502,6 +502,25 @@ Window.create = function(id, opts) {
...
+            });
+        };
+
+        webContents.on('crashed', () => {
```
StevenEBarbaro: Do we need to detach the events?

src/browser/api/window.js
joneit: @rdepena Adding to app event as `name`.... or should it be `windowName` ?

src/browser/api/window.js
```diff
@@ -502,6 +502,25 @@ Window.create = function(id, opts) {
             }
         });
 
+        var emitToAppAndWin = (...types) => {
```
StevenEBarbaro: These are only supposed to emit as the App when it's a main-window

joneit: I'll but a check before triggering the event routed _to_ the app. Is that what you mean?

joneit:
Manual test passed.

joneit:
@whyn07m3 so what folder in bitbucket for manual test app? it's a complex repo derived from _hello, openfin!_ app. Where do I put manual instructions?

HarsimranSingh:
@joneit check out https://bitbucket.org/openfin/test_apps/src/3c024a1f503d15944d435fa165faa2b773f0a183/apps/standard/?at=master

joneit:
Added 4 test runner tests, all named starting with `new Window (multiple preload scripts`, all public, and all with minimum version = `7.53.23.1` — so they won't run yet, but I have tested them with `alpha` (`7.53.22.11`) and they all pass.

src/browser/api/window.js
```diff
@@ -502,6 +502,28 @@ Window.create = function(id, opts) {
             }
         });
 
+        const emitToAppAndWin = (...types) => {
+            let isMainWindow = (uuid === name);
+            types.forEach(type => {
+                ofEvents.emit(route.window(type, uuid, name), { topic: 'window', type, uuid, name });
+                if (isMainWindow) {
+                    ofEvents.emit(route.application(`window-${type}`, uuid), { topic: 'application', type, uuid, name });
```
datamadic: My understanding here is that if it is the main window that it should propagate as an application event, if it is a child window it should be with the `window-` prefix. Is that correct? @rdepena @StevenEBarbaro @joneit

joneit: I'd like to put together an "event traffic" internal wiki to reduce the mystery...

joneit: @datamadic @rdepena ok, hoping (865c44f) addresses all concerns. 😖  Notwithstanding, note that I left `name` off the event object for the additional Application event. *_Is this correct?_*

src/browser/api/window.js
```diff
@@ -502,6 +502,35 @@ Window.create = function(id, opts) {
             }
         });
 
+        const emitToAppAndWin = (...types) => {
+            let isMainWindow = (uuid === name);
```
nicholasdgoodman: Are we _still_ doing this?

src/browser/api/window.js
```diff
@@ -477,7 +477,7 @@ Window.create = function(id, opts) {
             });
 
             // can't unhook when the 'closed' event fires; browserWindow is already destroyed then
-            browserWindow.webContents.removeAllListeners('page-favicon-updated');
+            webContents.removeAllListeners('page-favicon-updated');
```
rdepena: @joneit did you confirm that this event is handled with this change?

datamadic: ➕ ➕ 

@joneit when you have a sec can you review @rdepena's concern, the 6 and 7 hotfixes rely on this

src/browser/api/window.js
joneit: yeah i agree as a first step there should be one centralized `isMainWindow` core method even if, today, it still uses `uuid === name`... not sure if this is _that_ ticket though...?

src/browser/api/window.js
joneit: @rdepena no i did not; not worth the effort to test it; I'll just change it back. Point taken!

joneit:
by the way, i've prepped a v6 PR (#101), up to date with this PR.

Proof by diff of diffs:
```bash
git diff c589e0d RUN-3145_Emit-diagnostic-events-from-Window > 7.patch
git diff upstream/6.49.21/staging RUN-3145_v6_Emit-diagnostic-events-from-Window > 6.patch
diff 6.patch 7.patch
```
The resulting diff shows no differences (other than an extra blank line in v6 version).

rdepena:
needs tests 🌮 

Can we add the test runner links for these changes?

joneit:
There are no automated tests (that we can find); I am putting my manual test in bitbucket, with instructions.

rdepena:
@joneit this should be testable, Create app, wrap and listen to crash event then executeJavascript something that fails.

HarsimranSingh:
Merged into `7.53.22/staging` and `develop`

PR #101, opened 2017-07-19
RUN-3145 Emit diagnostic events from Window

joneit:
This PR is pending approval of #91 (v7 version).


joneit:
@StevenEBarbaro I made this PR against `6.49.21/staging` ... is that right?

PR #103, opened 2017-07-24
RUN-3165 remove reliance on the renderer for unload events 

datamadic:
This removes the need to listen for unload events on the renderer side. This is paired with [this runtime PR](https://github.com/openfin/runtime/pull/532). Listening via remote for the unload event was causing a browser side error when the child was cross origin. 

Tests will be posed after getting an upstream runtime to target. Local tests were at least as good as current stable / 22 though I get black screen crashes on almost every run for both stable and my locally built changes.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -356,9 +356,17 @@ Window.create = function(id, opts) {
         browserWindow.hide();
     };
 
-    function onDidUnload() {
+    const ofUnloadedHandler = (url, isMainFrame) => {
+
+        if (isMainFrame) {
+            ofEvents.emit(route.window('unload', uuid, name, false), identity);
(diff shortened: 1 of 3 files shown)


src/browser/api/window.js
```diff
@@ -416,6 +424,7 @@ Window.create = function(id, opts) {
         name = _options.name;
 
         const WINDOW_DOCUMENT_LOADED = 'document-loaded';
+        const OF_WINDOW_UNLOADED = 'of-window-unloaded';
```
rdepena: this event is a new event from the runtime?

rdepena:
👍

src/browser/api/window.js
datamadic: yeah its (this)[https://github.com/openfin/runtime/pull/533/files]

wenjunche:
👍

PR #106, opened 2017-07-25
RUN-3167 Socket server fix

rdepena:
Fixed issue where if the port was already in use we would not start the socket server.

#104 into 7/Staging

@datamadic @StevenEBarbaro 

Tests Pending


wenjunche:
👍

PR #107, opened 2017-07-25
RUN-3165 remove reliance on the renderer for unload events

datamadic:
This represents the hotfix changes bound for the 7.53.21.6 patch.

Requires runtime update ([this change set](https://github.com/openfin/runtime/pull/533) to hotfix)
@HarsimranSingh where should I make that to?


wenjunche:
👍

PR #126, opened 2017-08-02
RUN-3158 preload state will be stored in core state

whyn07m3:
ℹ️  This PR sets up the stage for subsequent preload script state changes.
State of the preload script will be located in OpenFinWindow object in the core state.
The list of available states shown here will be documented in JavaScript adapter and paired up with another PR.

---

✅  **Test results**
*Note: 2 new fails you see also happen when running against the latest develop. This PR does not introduce new regressions. These errors are likely due to runtime and core versions mismatch.*
• [Windows 7]()
• [Windows 10]()

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -763,6 +763,14 @@ Window.create = function(id, opts) {
         _window: browserWindow
     };
 
+    // Set initial preload state if window has preload script
+    if (_options.preload) {
+        winObj.preloadState = {
+            src: _options.preload,
+            state: 'idle'
+        };
+    }
+
     if (!coreState.getWinObjById(id)) {
         coreState.setWindowObj(id, winObj);
 
--- a/src/shapes.ts
+++ b/src/shapes.ts
@@ -91,6 +91,16 @@ export interface OpenFinWindow {
     id: number;
     name: string;
     uuid: string;
+    preloadState?: { // set if window has a preload script
+        src: String;
+        state:
+            'idle'| // initial state, not emitted, really here for internal use
+            'load-started'| // started loading/downloading the script file
+            'load-failed'| // failed to load/download the script file
+            'load-succeeded'| // preload script's content is loaded and ready to be eval'ed
+            'failed'| // preload script failed to eval
+            'succeeded'; // preload script eval'ed successfully
+    }[];
 }
 
 export interface BrowserWindow {


src/shapes.ts
```diff
@@ -91,6 +91,16 @@ export interface OpenFinWindow {
     id: number;
     name: string;
     uuid: string;
+    preloadState?: { // set if window has a preload script
+        src: String;
+        state:
+            'idle'| // initial state, not emitted, really here for internal use
```
HarsimranSingh: Super :neckbeard: but how do you feel about `'initial'` instead of `'idle'`? Just a thought

src/shapes.ts
```diff
@@ -91,6 +91,16 @@ export interface OpenFinWindow {
     id: number;
     name: string;
     uuid: string;
+    preloadState?: { // set if window has a preload script
+        src: String;
```
joneit: `src` should be [`PreloadOption`](https://github.com/HadoukenIO/core/blob/develop/src/browser/preload_scripts.ts#L28)

While you're at it, please change `PreloadOption` to a union type including string:
```js
type PreloadOption = string | PreloadInstance[];
```

joneit: do we really need `src` anyway? will `src` ever be anything other than the preload option? If not, isn't the option always avaiable via [`getOptions`](https://github.com/HadoukenIO/core/blob/develop/src/browser/api/window.js#L1062)?

src/shapes.ts
whyn07m3: @HarsimranSingh i'm okay with 'initial', the reason for 'idle': - script is in the queue/list and is 'idle', just sits there, maybe will be picked up later to download and execute maybe not, who knows? just being idle there...

@datamadic @rdepena @StevenEBarbaro @joneit  what are your thoughts on this/these?

joneit: `'initial'`

src/browser/api/window.js
```diff
@@ -763,6 +763,14 @@ Window.create = function(id, opts) {
         _window: browserWindow
     };
 
+    // Set initial preload state if window has preload script
+    if (_options.preload) {
+        winObj.preloadState = {
+            src: _options.preload,
```
joneit: Why not `url` instead of `src`? I realize `src` is used in the DOM, but for consistency with _our_ usage elsewhere in core when we have URL string.

src/shapes.ts
```diff
@@ -91,6 +91,16 @@ export interface OpenFinWindow {
...
+        state:
+            'idle'| // initial state, not emitted, really here for internal use
+            'load-started'| // started loading/downloading the script file
```
joneit: The state string is fine, but in reference to you comment: I think by "loading" here you mean when it's a `file://` scheme. However, I am using the term "load" differently, as in "fetch and load" where "fetch" gets the script into the cache (regardless of scheme); and "load" reads the cached file into memory. For consistency's sake, can we just say "fetch" instead of "load/download" ?

PR #128, opened 2017-08-02
RUN-3124 emit preload state events

whyn07m3:
ℹ️ This PR updates preload script states and emits relevant events.

🔗 **Paired** with [this PR](https://github.com/openfin/javascript-adapter/pull/317) in JavaScript adapter

---

**No test results** because this PR depends on [this previous PR](https://github.com/HadoukenIO/core/pull/126). Will run through tests once the other PR is consumed.

---

➕ **New tests:**
**preload-state-changed**
• [Window.addEventListener (preload-state-changed)](https://testing-dashboard.openfin.co/#/app/tests/597635eda3c7663182518fbf/edit)
• [Window.addEventListener (preload-state-changed) T2](https://testing-dashboard.openfin.co/#/app/tests/5976420ba3c7663182518fc1/edit)
• [Window.addEventListener (preload-state-changed) T2 (load-failed)](https://testing-dashboard.openfin.co/#/app/tests/59765d20a3c7663182518fc3/edit)
• [Window.addEventListener (preload-state-changed) T2 (failed)](https://testing-dashboard.openfin.co/#/app/tests/59766156a3c7663182518fc4/edit)
**preload-state-changing**
• [Window.addEventListener (preload-state-changing)](https://testing-dashboard.openfin.co/#/app/tests/597641d1a3c7663182518fc0/edit)
• [Window.addEventListener (preload-state-changing) T2](https://testing-dashboard.openfin.co/#/app/tests/59765835a3c7663182518fc2/edit)
• [Window.addEventListener (preload-state-changing) T2 (load-failed)](https://testing-dashboard.openfin.co/#/app/tests/59766206a3c7663182518fc6/edit)
• [Window.addEventListener (preload-state-changing) T2 (failed)](https://testing-dashboard.openfin.co/#/app/tests/597661cda3c7663182518fc5/edit)

PR #131, opened 2017-08-02
RUN-3125 added ability to query preload state

whyn07m3:
ℹ️  Adds ability to query for preload state by retrieving window info

🔗 **Paired** with [this PR](https://github.com/openfin/javascript-adapter/pull/318) which documents new window info payload

---

➕ New test:
• [Window.getInfo T2 (preloadState)](https://testing-dashboard.openfin.co/#/app/tests/59774ef9a3c7663182518fca/edit)

---

Local tests pass. **No public test results** because this PR depends on the [previous PR](https://github.com/HadoukenIO/core/pull/128). Will post them after previous PR is consumed.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1029,15 +1029,18 @@ Window.getGroup = function(identity) {
 
 
 Window.getWindowInfo = function(identity) {
-    let browserWindow = getElectronBrowserWindow(identity, 'get info for');
-    let webContents = browserWindow.webContents;
-
-    return {
-        url: webContents.getURL(),
-        title: webContents.getTitle(),
+    const browserWindow = getElectronBrowserWindow(identity, 'get info for');
+    const openfinWindow = Window.wrap(identity.uuid, identity.name);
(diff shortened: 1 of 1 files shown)


src/browser/api/window.js
```diff
@@ -1029,15 +1029,21 @@ Window.getGroup = function(identity) {
...
+    const windowInfo = {
+        canNavigateBack: webContents.canGoBack(),
         canNavigateForward: webContents.canGoForward(),
-        canNavigateBack: webContents.canGoBack()
+        title: webContents.getTitle(),
+        url: webContents.getURL()
     };
+
+    if (openfinWindow.preloadState) {
+        windowInfo.preloadState = openfinWindow.preloadState;
```
HarsimranSingh: Should preload state always be included in the payload? Is there a default value that could represent this? `null`?

whyn07m3: Good point, it can be anything, `null` or maybe `[]` (for easier user consumption). Just a matter of us deciding on what we want it to be

PR #138, opened 2017-08-11
RUN-3234 Preload fetch may fulfill early

joneit:
Regression [**tests**](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/598dd98bd2a02971da3a62be) passed (no relevant fails).

I simulated a fetch failure on my local to demonstrate the problem which resolved after fix applied.

### DIFF
--- a/src/browser/preload_scripts.ts
+++ b/src/browser/preload_scripts.ts
@@ -35,7 +35,7 @@ interface PreloadInstance {
 interface PreloadFetched extends PreloadInstance {
     scriptPath: string; // location on disc of cached fetch
 }
-type FetchResolver = (value: PreloadInstance | PreloadFetched | FetchResponse) => void;
+type FetchResolver = (value?: PreloadInstance | PreloadFetched | FetchResponse) => void;
 type Resolver = (value?: any) => void;
 type Rejector = (reason?: Error) => void;
 
@@ -113,7 +113,7 @@ export function fetchAndLoadPreloadScripts(
 
 
 // resolves to type `PreloadFetched` on success
-// resolves to the input (type `PreloadInstance`) when asset fails to be cached
+// resolves to `undefined` when fetch fails to download asset to Chromium cache
 function fetch(identity: Identity, preloadScript: PreloadInstance): Promise<FetchResponse> {
     return new Promise((resolve: FetchResolver, reject: Rejector) => {
         cachedFetch(identity.uuid, preloadScript.url, (fetchError: null | Error, scriptPath: string | undefined) => {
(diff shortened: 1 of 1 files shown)


src/browser/preload_scripts.ts
```diff
@@ -113,29 +113,29 @@ export function fetchAndLoadPreloadScripts(
...
+// resolves to `undefined` when fetch fails to download asset to Chromium cache
 function fetch(identity: Identity, preloadScript: PreloadInstance): Promise<FetchResponse> {
     return new Promise((resolve: FetchResolver, reject: Rejector) => {
         cachedFetch(identity.uuid, preloadScript.url, (fetchError: null | Error, scriptPath: string | undefined) => {
             if (!fetchError) {
                 resolve({identity, preloadScript, scriptPath});
             } else {
                 updatePreloadState(identity, preloadScript, 'load-failed');
-                reject();
+                resolve();
```
datamadic: was this initally just a bug (`reject` to `resolve`)?

joneit: well, no, not originally, because originally we were using a different paradigm.

### Originally we had this:
1. Fetch (to Chromium cache) & load (from Chromium cache) all preload scripts via resource fetcher _for each and every page load_
2. Wait for all to complete loading
3. Proceed to `loadURL` _if and only if_ all "required" (not optional) scripts loaded successfully
4. Eval loop in api_decorator executes all available scripts
5. Scripts are released from memory after eval

So `reject` was being called on `if (!optional)` when a required script failed to load, and this call was an overlooked remnant from that code.

### We then changed to this:
1. Fetch & load all of them via resource fetcher _if and only if_ not previously fetched, caching scripts in memory (as a side-effect) as they are successfully loaded
3. Wait for any new scripts to complete loading
4. Proceed to `loadURL` regardless of load success
5. Eval loop in api_decorator executes all successfully loaded scripts _if and only if_ that set includes all required scripts
6. Scripts remain in memory for subsequent page loads (any app, any window)

#### Advantages of new paradigm:
1. Our resource fetcher currently not smart enough to skip re-fetching external files over and over again, so maintaining an in-memory cache gives us simple control of that
2. Skips reloading locally from disc cache on every page (minor impact though that is)

#### Disadvantage of new paradigm:
1. Uses more memory (for scripts), judged to be minor impact
2. Edited scripts are not refetched until runtime restarts

We made one final change to the above when Omer pointed out that local files (`file://`) are never actually fetched, but load directly from their native disc locations. #147 changes the logic to always reload local files so app developers can freely edit them and expect to see different results without restarting runtime (however see #147 for limitations on this).

joneit:
Rebased to current `canary` (`8.56.24.3`).

src/browser/preload_scripts.ts
```diff
@@ -113,29 +113,29 @@ export function fetchAndLoadPreloadScripts(
...
-// resolves to the input (type `PreloadFetched`) when fetched asset fails to load (can't be read from cache)
+// resolves to `undefined` when above fetch failed or when successfully fetched asset fails to load from Chromium cache
 function load(opts: FetchResponse): Promise<undefined> {
     return new Promise((resolve: Resolver, reject: Rejector) => {
-        const {identity, preloadScript, scriptPath} = opts;
-
-        if (!scriptPath) {
-            resolve(); // bad fetch but we only get this far if script was optional so resolve
+        if (!opts || !opts.scriptPath) {
+            resolve(); // got fetchError above OR no error but no scriptPath either; in any case don't attempt to load
```
rdepena: hmm, not sure this should resolve. I get that we need to continue but this function viewed in isolation should fail under the two stated circumstances.

joneit: We are caching the scripts as a side-effect, the call to `System.setPreloadScript(preloadScript.url, scriptText);`. (We could make this clearer in a comment at top.) Point is resolving early skips the caching call.

whyn07m3:
Also, as a reminder, this needs a second batch of tests (after rebasing)

joneit:
I actually updated the "tests" link in the description at top, which I admit is unclear; I will put it in rebase comment in future.

PR #141, opened 2017-08-15
RUN-3260: added method to get app asset info

weiteho:
added method to allow users to get existing app asset info

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -243,6 +243,12 @@ exports.System = {
     getAllApplications: function() {
         return coreState.getAllApplications();
     },
+    getAppAssetInfo: function(identity, options, callback, errorCallback) {
+        var appObject = coreState.getAppObjByUuid(identity.uuid);
+        options.srcUrl = (appObject || {})._configUrl;
+        var appAssetsFetcher = require('../rvm/runtime_initiated_topics/app_assets').appAssetsFetcher;
+        appAssetsFetcher.fetchAppAsset(options.srcUrl, options.alias, callback, errorCallback);
+    },
     getCommandLineArguments: function() {
         return electronApp.getCommandLineArguments();
     },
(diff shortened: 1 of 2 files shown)


src/browser/api_protocol/api_handlers/system.js
```diff
@@ -88,6 +89,18 @@ function SystemApiHandler() {
...
+    function getAppAssetInfo(identity, message, ack, nack) {
+        let options = message.payload;
+
+        System.getAppAssetInfo(identity, options, function(data) {
+            let dataAck = _.clone(successAck);
+            dataAck.data = data;
+            ack(dataAck);
+        }, function(err) {
+            nack(err);
+        });
```
joneit: Just reference `nack` here instead of `function(err) { nack(err); }`

src/browser/api/system.js
```diff
@@ -243,6 +243,12 @@ exports.System = {
     getAllApplications: function() {
         return coreState.getAllApplications();
     },
+    getAppAssetInfo: function(identity, options, callback, errorCallback) {
```
joneit: Question: Wondering why this isn't `Application.getAssetInfo` instead?

joneit:
Note sister PR openfin/javascript-adapter#324

wenjunche:
talked to @StevenEBarbaro:  path should not be included in the response for security.

wenjunche:
This needs to go to 7.53.23

src/browser/api/system.js
```diff
@@ -243,6 +243,11 @@ exports.System = {
...
+        options.srcUrl = coreState.getConfigUrlByUuid(identity.uuid);
+        var appAssetsFetcher = require('../rvm/runtime_initiated_topics/app_assets').appAssetsFetcher;
```
HarsimranSingh: Can this be moved to the top of the file?

PR #142, opened 2017-08-16
RUN-2974 default backgroundColor to #0FFF when transparent is true

HarsimranSingh:
Related to this PR: https://github.com/openfin/runtime/pull/580

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -269,6 +269,10 @@ module.exports = {
             newOptions.preload = options.preload || [];
         }
 
+        if (newOptions.transparent) {
+            newOptions.backgroundColor = '#0FFF';
+        }
+
         if (returnAsString) {
             return JSON.stringify(newOptions);
         } else {


src/browser/convert_options.js
```diff
@@ -269,6 +269,10 @@ module.exports = {
             newOptions.preload = options.preload || [];
         }
 
+        if (newOptions.transparent) {
+            newOptions.backgroundColor = '#0FFF';
```
joneit: would `'transparent'` work here?

if so I wonder if we really even need `options.transparent'`⁉

TomaszMa: @joneit I don't know if we can change 'backgroundColor' later after window will be created, probably not (80%), this is question to @HarsimranSingh . As far as I know 'backgorundColor' is openfin config .json variable to set up application window background (I am new in openfin and not everything is clear for me, e.g.: if we has access to config.json variables from java script level). Problem was that when you set 'transparent' to true in config.json and leave 'backgorundColor' default nothing will change for user, but in runtime code window will be able to support alpha chanel. Thats why we changed default background color to something with alpha 0 when 'transparent' is set to true. Of curse user can override backgroundColor in config.json to something opaque or semi-transparent.

HarsimranSingh:
Closing for now. I'll open it again when transparency is ready to be merged.

PR #146, opened 2017-08-18
RUN-3251 feature/toggle log levels

datamadic:
Adds the ability to toggle the verbosity levels programmaticly.

Needs [adapter PR](https://github.com/openfin/javascript-adapter/pull/328)
Needs [runtime PR](https://github.com/openfin/runtime/pull/582)

[tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599668e5d2a02971da3a62e7) (fails passed in isolation / parody with vanilla build) 
[added tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59966fd0d2a02971da3a62ed)

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -39,18 +39,22 @@ import ofEvents from '../of_events';
 const ProcessTracker = require('../process_tracker.js');
 import route from '../../common/route';
 import { fetchAndLoadPreloadScripts } from '../preload_scripts';
-
+const verbose = 'verbose';
+const info = 'info';
+const warning = 'warning';
+const error = 'error';
+const fatal = 'fatal';
 const logLevelMappings = new Map([
-    ['verbose', -1],
-    ['info', 0],
-    ['warning', 1],
-    ['error', 2],
-    ['fatal', 3],
-    [-1, 'verbose'],
-    [0, 'info'],
-    [1, 'warning'],
-    [2, 'error'],
-    [3, 'fatal']
+    [verbose, -1],
+    [info, 0],
+    [warning, 1],
+    [error, 2],
+    [fatal, 3],
+    [-1, verbose],
+    [0, info],
+    [1, warning],
+    [2, error],
+    [3, fatal]
 ]);
 
 const defaultProc = {
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -40,6 +40,18 @@ const ProcessTracker = require('../process_tracker.js');
 import route from '../../common/route';
 import { fetchAndLoadPreloadScripts } from '../preload_scripts';
 
+const logLevelMappings = new Map([
+    ['verbose', -1],
+    ['info', 0],
+    ['warning', 1],
+    ['error', 2],
+    ['fatal', 3],
+    [-1, 'verbose'],
+    [0, 'info'],
+    [1, 'warning'],
+    [2, 'error'],
+    [3, 'fatal']
+]);
 
 const defaultProc = {
     getCpuUsage: function() {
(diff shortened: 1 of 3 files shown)


src/browser/api/system.js
```diff
@@ -39,7 +39,23 @@ import ofEvents from '../of_events';
 const ProcessTracker = require('../process_tracker.js');
 import route from '../../common/route';
 import { fetchAndLoadPreloadScripts } from '../preload_scripts';
-
+const verbose = 'verbose';
+const info = 'info';
+const warning = 'warning';
+const error = 'error';
+const fatal = 'fatal';
+const logLevelMappings = new Map([
```
joneit: In spirit of KISS, why not an array with `findIndex` method to get level? I guess it's that `-1`? But was `-1` really necessary?

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -41,6 +41,7 @@ function SystemApiHandler() {
         'get-el-ipc-config': getElIPCConfig,
         'get-environment-variable': { apiFunc: getEnvironmentVariable, apiPath: '.getEnvironmentVariable' },
         'get-host-specs': { apiFunc: getHostSpecs, apiPath: '.getHostSpecs' },
+        'get-min-log-level': getMinLogLevel,
```
joneit: As mentioned in adapter PR, my feeling is that `getMaxLogLevel`/`setMaxLogLevel` might read with more meaning. As in "this is the _most_ detail we will log"... this means reversing the numbering of the levels. Am I the only one looking at it this way?

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -71,6 +73,35 @@ function SystemApiHandler() {
 
     apiProtocolBase.registerActionMap(SystemApiHandlerMap, 'System');
 
+    function didFail(e) {
+        return e !== undefined && e.constructor === Error;
+    }
+
+    function setMinLogLevel(identity, message, ack, nack) {
+        const { payload: { level } } = message;
+        const response = System.setMinLogLevel(level);
```
joneit: while i like the destructuring, wouldn't it just be simpler to say:
```js
    const response = System.setMinLogLevel(message.payload.level);
```

src/browser/api/system.js
datamadic: Because this becomes a native call, I juts mirrored the levels as that is what you have to pass anyway, using the array index would be a bit clunky with the `index - 1` what would have to be used

src/browser/api_protocol/api_handlers/system.js
datamadic: I think min is more natural, what are the chances that you want to set `info` log levels and not get 'error' as well?

src/browser/api_protocol/api_handlers/system.js
datamadic: six of one, half dozen of the other...

src/browser/api_protocol/api_handlers/system.js
nicholasdgoodman: Not to confuse the issue, but IMO neither terminology (min vs max) is appropriate here.

Compare to Java _log4j_:

```
log.setLevel(Level.WARN);
```

Or MongoDB:

```
db.setLogLevel(<level>, <component>)
```

Googling `'setMinLogLevel'` finds that Chromium is the only major project to use this terminology. Meanwhile `'setMaxLogLevel'` appears to have no common usage.

src/browser/api/system.js
whyn07m3: In another PR, this will need to be required in another file, so this needs this to be exported

whyn07m3: Since these are log levels, and they will be reused elsewhere, should we put them into `log.ts`?

PR #147, opened 2017-08-21
RUN-3275 Always re/fetch local preload script files

joneit:
\[See sister [PR#330](https://github.com/openfin/javascript-adapter/pull/330) in openfin/javascript-adapter.]
 
### This PR:

@omerli points out local files normally bypass Chromium cache anyway and therefore _would always be reloaded on every page;_ and to aid app development, we should do the same. 

1. **Adjust logic to always reload local files** on `new Window({preload:'file://...'})`, even if already loaded by a previously created window.<br>
_Specifically:_ Local files (resources with `file://` scheme) are now _excepted_ from our _temporary_ work-around that generally _avoids_ re-fetching/loading previously loaded preload scripts. (Workaround to be removed by [RUN-3227](https://appoji.jira.com/browse/RUN-3227) once [RUN-3162](https://appoji.jira.com/browse/RUN-3162) is resolved.)<br>
_Caveat:_ Normal page refresh does _not_ reload local files, only on `new Window`.<br><br>
2. **Add a `preload` option** to `fin.desktop.System.clearCache`.<br>
So now app developer can type `fin.desktop.System.clearCache({preload:true})` into console and then reload page to refetch a (presumably modified) preload scripts.

### Tests:
Tested locally. I could add my tests for both items above to test runner's manual test suite though not sure if its worth the effort because:
* test would require a lot of effort (would have to rewrite script files) 
* these were simple changes already tested manually 
* item 1 above is temporary until better resource fetcher comes along

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -74,7 +74,8 @@ const defaultProc = {
     }
 };
 
-const preloadScriptsCache = {};
+let preloadScriptsCache;
+clearPreloadCache();
 
 let MonitorInfo;
 let Session;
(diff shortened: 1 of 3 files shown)


joneit:
Rebased to current `canary` (`8.56.24.3`).

joneit:
Rebased to current `canary` (`runtime-v8.56.24.10`).
[Retested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59a2fd71d2a02971da3a6364) clean.

rdepena:
👍

joneit:
@whyn07m3 related the following to me verbally and has no opinion about merging it as is:
1. if we have `clearCatch({preload:true})` then we don’t also need `file://` check, which is a patch on a patch and will be removed eventually anyway
2. `file://` should more properly be `file:///`

joneit:
@whyn07m3 :
1. `file://` check is better user (developer) experience because without this developers need to know to call `clearCache` from console if they expect to see their updated scripts executed. That said, I don't mind removing it if others feel the same way.
2. The proper "scheme" is `file://` because `file:///` implies the default (`file://localhost/`) which could conceivably be given

PR #148, opened 2017-08-21
RUN-2760 Support deep linking on app launch and when already running

VanessaSergeon:
Deep linking

[Issue discussion](https://github.com/openfin/Internal-Wiki/issues/6)

[Tests windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599b1e0cd2a02971da3a6302)

[Tests windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599b2288d2a02971da3a6307)

3 tests fail. 2 pass in isolation. 1 fails in alpha as well.

Currently there are no tests specifically for deep linking. Will Make a manual test.

[Docs](https://github.com/openfin/javascript-adapter/pull/329)


src/browser/api/application.js
```diff
@@ -438,14 +438,14 @@ Application.revokeWindowAccess = function() {
     console.warn('Deprecated');
 };
 
-Application.run = function(identity, configUrl = '') {
+Application.run = function(identity, userAppConfigArgs, configUrl = '') {
```
joneit: Generally speaking, new params must be added to end of existing param list to maintain backwards compatibility

src/browser/api/application.js
```diff
@@ -417,7 +417,7 @@ Application.restart = function(identity) {
 
     try {
         Application.close(identity, true, () => {
-            Application.run(identity, appObj._configUrl);
+            Application.run(identity, undefined, appObj._configUrl);
```
joneit: `, undefined` would not be necessary if new param were at end

src/browser/api/application.js
VanessaSergeon: Preferably yes. Had done it that way to avoid having to set a default but will just set the default to undefined.

index.js
```diff
@@ -504,6 +504,10 @@ function launchApp(argo, startExternalAdapterServer) {
             configObject: { licenseKey }
         } = configuration;
 
+        if (argo['user-app-config-args']) {
+            configObject['startup_app'].url = `${configObject['startup_app'].url}?${argo['user-app-config-args']}`;
```
joneit: I think we need to accommodate original `url` possibly already having a querystring by appending with `&` rather than `?`, something like :
```js
var options = configObject['startup_app'];
configObject['startup_app'].url += options.url.indexOf('?') < 0 ? '?' : '&';
options.url += argo['user-app-config-args'];
```
Because multiple `?` would be unexpected.
This does raise the question however as to whether RVM `$$` params should _overwrite_ existing url params or should just be appended as duplicate keys (permitted in urls).

VanessaSergeon: Thats a great point. Updating and testing it now.

joneit:
I’m thinking prefix `of$` would be more unique, more meaningful, and more recognizable than `$$`. 

Omer's response:
> don't have a strong opinion about it - the only reason I went with 2 dollar signs is to prevent a possible collision if someone decided that their query string params should start with $
>
> if everyone is ok with that, we can change it in the RVM

@rdepena @StevenEBarbaro @HarsimranSingh Any objections?

rdepena:
@joneit @omerli I don't have any strong opinions on this either, its the type of thing you just pick and learn, of$ or $$ really don't make a difference when you are using the feature.

HarsimranSingh:
@VanessaSergeon update the title so it's in the format `RUN-#### <commit message>`

PR #152, opened 2017-08-23
RUN-3136 Limit accelerator registration per options

joneit:
This PR only registers "shortcut" keys that are needed by the page.

Base on current `canary` (`8.56.24.3`).

[Tested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599ed18cd2a02971da3a633f) clean.

#### Registering accelerator shortcuts

Previously designated "shortcut" keys were always registered as accelerators, although their handlers only took action as needed. This defect pre-empted the DOM events from firing for those keys so registered.

This improvement allows DOM event handler/listeners to catch these keys in `keydown` handler —
 so long as they are not required per `options.accelerator`. This includes **F5**.

```js
function f5handler(e) {
    if (e.keyCode === 116) console.log('F5 pressed');
}

document.addEventListener('keydown', f5handler); // "listener"
// or:
document.onkeydown = f5handler; // "handler"
```

Documented in a new internal [wiki](https://github.com/openfin/Internal-Wiki/wiki/Trapping-F5-keypress).

#### jshint change

This PR also changes the jshint "no unused vars" rule to ignore unused formal params (in function definitions). I feel it is more helpful to keep unused formal params in place, especially for functions with mandated calling signatures, such as event handler, even when such vars are unused therein. Thoughts?

### DIFF
--- a/index.js
+++ b/index.js
@@ -610,15 +610,15 @@ function registerShortcuts() {
         const webContents = browserWindow.webContents;
 
         if (accelerator.zoom) {
-            const zoom = zoomIn => { return () => { webContents.send('zoom', zoomIn, zoomIn === undefined); }; };
+            const zoom = zoomIncrement => { return () => { webContents.send('zoom', zoomIncrement); }; };
 
-            globalShortcut.register('CommandOrControl+0', zoom(undefined));
+            globalShortcut.register('CommandOrControl+0', zoom(0));
 
-            globalShortcut.register('CommandOrControl+=', zoom(true));
-            globalShortcut.register('CommandOrControl+Plus', zoom(true));
+            globalShortcut.register('CommandOrControl+=', zoom(+1));
+            globalShortcut.register('CommandOrControl+Plus', zoom(+1));
 
-            globalShortcut.register('CommandOrControl+-', zoom(false));
-            globalShortcut.register('CommandOrControl+_', zoom(false));
+            globalShortcut.register('CommandOrControl+-', zoom(-1));
+            globalShortcut.register('CommandOrControl+_', zoom(-1));
         }
 
         if (accelerator.devtools) {
--- a/index.js
+++ b/index.js
@@ -604,90 +604,43 @@ function initFirstApp(configObject, configUrl, licenseKey) {
 }
 
 function registerShortcuts() {
-    const resetZoomShortcut = 'CommandOrControl+0';
-    const zoomInShortcut = 'CommandOrControl+=';
-    const zoomInShiftShortcut = 'CommandOrControl+Plus';
-    const zoomOutShortcut = 'CommandOrControl+-';
-    const zoomOutShiftShortcut = 'CommandOrControl+_';
-    const devToolsShortcut = 'CommandOrControl+Shift+I';
-    const reloadF5Shortcut = 'F5';
-    const reloadShiftF5Shortcut = 'Shift+F5';
-    const reloadCtrlRShortcut = 'CommandOrControl+R';
-    const reloadCtrlShiftRShortcut = 'CommandOrControl+Shift+R';
+    app.on('browser-window-focus', (event, browserWindow) => {
+        const windowOptions = coreState.getWindowOptionsById(browserWindow.id);
+        const accelerator = windowOptions && windowOptions.accelerator || {};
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -595,90 +595,43 @@ function initFirstApp(configObject, configUrl, licenseKey) {
...
+    app.on('browser-window-focus', (event, browserWindow) => {
+        const windowOptions = coreState.getWindowOptionsById(browserWindow.id);
+        const accelerator = windowOptions && windowOptions.accelerator || {};
+        const webContents = browserWindow.webContents;
 
-    let reload = () => {
-        let browserWindow = BrowserWindow.getFocusedWindow();
-        let windowOptions = browserWindow && coreState.getWindowOptionsById(browserWindow.id);
+        if (accelerator.zoom) {
+            const zoom = zoomIn => { return () => { webContents.send('zoom', zoomIn, zoomIn === undefined); }; };
```
HarsimranSingh: This can be refactored. Instead of sending zoom levels to the renderer to deal with, we can handle everything on the browser side by calling `browserWindow.webContents.getZoomLevel()` and `browserWindow.webContents.setZoomLevel()`.

https://github.com/HadoukenIO/core/blob/develop/src/renderer/api-decorator.js#L175

HarsimranSingh: This could also be moved to another ticket/PR. I'll leave it up to you.

joneit:
Simplified zoom funcs and rebased to `canary` (`runtime-v8.56.24.10`).

@sim as discussed we won't do your suggested refactor as the Chromium so-called "by design" behavior [is undesirable](https://bugs.chromium.org/p/chromium/issues/detail?id=390775).

index.js
joneit: As discussed we won't do your suggested refactor as the Chromium so-called "by design" behavior [is undesirable](https://bugs.chromium.org/p/chromium/issues/detail?id=390775).

src/renderer/api-decorator.js
```diff
@@ -172,13 +172,8 @@ limitations under the License.
...
         // messages sent from the browser process
-        ipc.on(`zoom-${renderFrameId}`, (event, zoomIn, reset) => {
-            if (reset) {
-                webFrame.setZoomLevel(0);
-            } else {
-                let level = Math.floor(webFrame.getZoomLevel());
-                webFrame.setZoomLevel(zoomIn ? ++level : --level);
-            }
+        ipc.on(`zoom-${renderFrameId}`, (event, zoomIncrement) => {
+            webFrame.setZoomLevel(zoomIncrement ? webFrame.getZoomLevel() + zoomIncrement : 0);
```
HarsimranSingh: Is there a reason for getting rid of `Math.floor()`? Usually when you get the zoom level it doesn't come back as an integer value.

joneit: yeah [it does](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#contentsgetzoomlevelcallback) (come back as a number)

joneit: @HarsimranSingh yeah [it does](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#contentsgetzoomlevelcallback)! (come back as a number)

joneit: Will add it back in.

joneit:
@HarsimranSingh `Window.setZoomLevel` now sends message to renderer because we want zoom to affect just this window and not all windows loaded from same domain as this window.

joneit:
Rebased to new `canary` (`8.56.24.14`).
Rerunnning regression/integration tests...

src/browser/api/window.js
```diff
@@ -1561,7 +1561,8 @@ Window.getZoomLevel = function(identity, callback) {
 Window.setZoomLevel = function(identity, level) {
     let browserWindow = getElectronBrowserWindow(identity, 'set zoom level for');
 
-    browserWindow.webContents.setZoomLevel(level);
+    // browserWindow.webContents.setZoomLevel(level); // zooms all windows loaded from same domain
```
VanessaSergeon: Is this commented out code intentional?

joneit: yes, as a caution to future maintainers not to revert to calling `setZoomLevel` directly, which gives the wrong result

PR #153, opened 2017-08-25
RUN-3286 App assets API does not work if not launched by manifest

wenjunche:
Test results:
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599f65a4d2a02971da3a6347
)
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/599f6542d2a02971da3a6346)

Also, fixed code in authorization.js according to [https://github.com/openfin/openfin/wiki/RVM-Protocol]().  created RUN-3290 to track the issue.

New versions of .net and Java adapters are required to test app asset APIs from external adapters.

### DIFF
--- a/src/browser/api_protocol/api_handlers/authorization.js
+++ b/src/browser/api_protocol/api_handlers/authorization.js
@@ -46,7 +46,7 @@ function registerExternalConnection(identity, message, ack) {
         token
     };
 
-    addPendingAuthentication(uuidToRegister, token, null, identity);
+    addPendingAuthentication(uuidToRegister, token, null, identity, null);
     ack(dataAck);
 }
 
@@ -74,7 +74,7 @@ function onRequestExternalAuth(id, message) {
     }
 
     // UUID assignment priority: mapped process, client-requested, then auto-generated
-    var uuid = (extProcess || {}).uuid || uuidRequested || electronApp.generateGUID();
+    const uuid = (extProcess || {}).uuid || uuidRequested || electronApp.generateGUID();
 
     if (pendingAuthentications.has(uuid)) {
         return;
@@ -83,7 +83,7 @@ function onRequestExternalAuth(id, message) {
     file = getAuthFile();
     token = electronApp.generateGUID();
 
-    addPendingAuthentication(uuid, token, file);
+    addPendingAuthentication(uuid, token, file, null, message.payload);
 
     socketServer.send(id, JSON.stringify({
         action: 'external-authorization-response',
@@ -99,7 +99,8 @@ function onRequestAuthorization(id, data) {
     const uuid = data.payload.uuid;
     const authObj = pendingAuthentications.get(uuid);
     const externalConnObj = Object.assign({}, data.payload, {
-        id
+        id,
+        configUrl: authObj.authReqPayload && authObj.authReqPayload.configUrl
     });
 
     //Check if the file and token were written.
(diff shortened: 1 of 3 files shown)


StevenEBarbaro:
👍

HarsimranSingh:
PR title format `RUN-#### <short description>` (no colon after the number)

src/browser/api_protocol/api_handlers/authorization.js
```diff
@@ -99,7 +99,8 @@ function onRequestAuthorization(id, data) {
     const uuid = data.payload.uuid;
     const authObj = pendingAuthentications.get(uuid);
     const externalConnObj = Object.assign({}, data.payload, {
-        id
+        id,
+        configUrl: authObj.authReqPayload && authObj.authReqPayload.configUrl
```
whyn07m3: This can maybe end up being `configUrl: null` or other falsy values. Will it brake somewhere down the line then? Maybe be should be this instead: 
```
const externalConnObj = Object.assign({}, data.payload, { id });

if (authObj.authReqPayload) {
   externalConnObj.configUrl = authObj.authReqPayload.configUrl;
}
```

src/browser/api_protocol/api_handlers/authorization.js
```diff
@@ -123,14 +124,14 @@ function onRequestAuthorization(id, data) {
 
         rvmMessageBus.registerLicenseInfo({
             data: {
-                licenseKey: externalConnObj.licenseKey,
-                client: externalConnObj.client,
+                licenseKey: authObj.authReqPayload && authObj.authReqPayload.licenseKey,
+                client: authObj.authReqPayload && authObj.authReqPayload.client,
```
whyn07m3: Same in these case as in the comment above...?

src/browser/core_state.ts
```diff
@@ -516,6 +521,15 @@ export function getAppAncestor(descendantAppUuid: string): Shapes.App {
     }
 }
 
+function getExternalAncestor(descendantAppUuid: string): any {
```
whyn07m3: Can we specify shape here instead of `any`?

src/browser/core_state.ts
```diff
@@ -211,7 +211,12 @@ export function getUuidBySourceUrl(sourceUrl: string): string|boolean {
 
 export function getConfigUrlByUuid(uuid: string): string|boolean {
     const app = getAppAncestor(uuid);
-    return app && app._configUrl;
+    if  (app && app._configUrl) {
+        return app._configUrl;
+    } else {
+        const externalApp = getExternalAncestor(uuid);
+        return externalApp && externalApp.configUrl;
```
whyn07m3: Allowed return values are `string|boolean`. This would have to be `return !!externalApp && externalApp.configUrl;`

whyn07m3:
Also, we need to test this

src/browser/core_state.ts
nicholasdgoodman: this function should not be returning `string|boolean` and falsey values should not be coerced to boolean. instead they should be coerced to `undefined` via:

```
// if getExternalAncestor is written correctly, this coercion isn't necessary.
const externalApp = getExternalAncestor(uuid) || {};

// returns a string value or undefined
return externalApp && externalApp.configUrl; 
```

src/browser/api_protocol/api_handlers/authorization.js
wenjunche: added check for authObj

src/browser/api_protocol/api_handlers/authorization.js
wenjunche: Added check for authObj

src/browser/core_state.ts
wenjunche: We need to create shapes for external connection

wenjunche:
Added following check in authenticateUuid

if (!authObj) {
        cb(false, 'Invalid UUID: ' + authRequest.uuid);

So onRequestAuthorization returns an error if there is no pending connection for the UUID.

Modified Java adapter to test the condition (sending request-authorization without request-external-authorization)

src/browser/core_state.ts
whyn07m3: I like this better 👍

whyn07m3: ... typescript reflects what we've had, and string|boolean is what we've had. Unless we start refactoring throughout, just safer to stick to what's there now?

PR #155, opened 2017-08-27
RUN-3161 Wait for ready-to-show

joneit:
I saved the state of `ready-to-show` event-fired directly on the `browserWindow` object (much as we save `_options` there). Seems like a good place for it inasmuch as it reflects `browserWindow` state. Nonetheless I throw it out there: Is there a better place for (what can also be thought of as) a window instance var?

@HarsimranSingh @StevenEBarbaro How do we test this? Is there already a v8 test for this (that tests the existing `setTimeout` solution)?

Based on current `canary` (`runtime-v8.56.24.10`).
Regression [tested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59a30434d2a02971da3a6365) clean.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -379,6 +379,11 @@ Window.create = function(id, opts) {
 
         browserWindow = BrowserWindow.fromId(id);
 
+        // set this listener up as soon as possible
+        browserWindow._whenReadyToShow = new Promise(ready => {
+            browserWindow.once('ready-to-show', ready);
+        });
+
         // this is a first pass at teardown. for now, push the unsubscribe
         // function for each subscription you make, on closed, remove them all
         // if you listen on 'closed' it will crash as your resources are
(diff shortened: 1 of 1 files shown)


joneit:
Rebased to `8.56.24.19`.
Retesting...

rdepena:
@joneit have you looked into @HarsimranSingh feedback?

src/browser/api/window.js
```diff
@@ -1708,22 +1708,26 @@ function getBoundsCacheSafeFileName(identity) {
 
 function applyAdditionalOptionsToWindowOnVisible(browserWindow, callback) {
     if (browserWindow.isVisible()) {
-        callback();
+        readySetGo();
```
HarsimranSingh: Actually, you can ignore this change. This condition, `if (browserWindow.isVisible())`, was only added because we reuse this function to reapply visibility-sensitive features, like rounded corners and resize regions, for `updateOptions()`. On window creation, which is the else statement, is the only point we care about `ready-to-show`.

HarsimranSingh:
Are you actually receiving the `ready-to-show` event? I can't seem to get it on app or child window creation through the API.

src/browser/api/window.js
joneit: So I think you're saying leave the conditional only call `callback` instead of `readSetGo`, right?

HarsimranSingh: right

joneit:
That last commit (3f6c311) per conversation with @HarsimranSingh that we should start listening for `ready-to-show` asap.

Rebased to `8.56.24.26` and [retested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59b022d1d2a02971da3a63c1).

PR #156, opened 2017-08-28
RUN-3127 Fix problem with "backgroundColor" window option.

TomaszMa:
This PR is part of fix for RUN-3127. See comment in code.

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -429,12 +429,9 @@ limitations under the License.
         options.openfin = true;
         options.preload = 'preload' in options ? options.preload : winOpts.preload;
         options.uuid = winOpts.uuid;
-        if (options.backgroundColor === undefined &&
-            winOpts.backgroundColor !== undefined) {
-          // Apply parent window background color to child window when child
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -429,6 +429,12 @@ limitations under the License.
         options.openfin = true;
         options.preload = 'preload' in options ? options.preload : winOpts.preload;
         options.uuid = winOpts.uuid;
+        if (options.backgroundColor === undefined &&
```
HarsimranSingh: @TomaszMa can you move this code here https://github.com/HadoukenIO/core/blob/develop/src/browser/api/window.js#L405

This is the unofficial place we've been using to handle the case where the child window should inherit an option from it's parent.

HarsimranSingh:
PR title format `RUN-#### <short description>` (no colon after the number)

src/renderer/api-decorator.js
TomaszMa: If I could, I would move the code where you proposed, but check need to be done before we convert options to electron options (which is done few lines below my changes).

PR #157, opened 2017-08-28
RUN-2905 win.show() restores maximized window

joneit:
Regression [test](https://en.wikipedia.org/wiki/File_URI_scheme)ed.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1373,6 +1373,12 @@ Window.show = function(identity, force = false) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/api/window.js
```diff
@@ -1373,6 +1373,12 @@ Window.show = function(identity, force = false) {
         return;
     }
 
+    if (browserWindow.isMaximized() && browserWindow.isVisible()) {
```
HarsimranSingh: Move this code into `defaultAction()`

joneit:
Also rebased to `canary` (`8.56.23.11`) on last commit.
Retesting now...

PR #160, opened 2017-08-30
Read javascript-adapter from dir and then asar; for ease of development

HarsimranSingh:
Tests came back clean:
[Win 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59a876f8d2a02971da3a639d)
[Win 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59a87c0bd2a02971da3a639f)

### DIFF
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -19,7 +19,19 @@ let fs = require('fs');
 let path = require('path');
 let me = fs.readFileSync(path.join(__dirname, 'api-decorator.js'), 'utf8');
 
-let jsAdapter = fs.readFileSync(path.join(process.resourcesPath, 'adapter.asar', 'openfin-desktop.js'), 'utf8');
+// check resources/adapter/openfin-desktop.js then
+// resources/adapter.asar/openfin-desktop.js
+// for ease of developement
+let jsAdapter = '';
+const searchPaths = ['adapter', 'adapter.asar'];
+for (let adapterPath of searchPaths) {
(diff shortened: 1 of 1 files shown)


src/renderer/main.js
```diff
@@ -19,7 +19,19 @@ let fs = require('fs');
 let path = require('path');
 let me = fs.readFileSync(path.join(__dirname, 'api-decorator.js'), 'utf8');
 
-let jsAdapter = fs.readFileSync(path.join(process.resourcesPath, 'adapter.asar', 'openfin-desktop.js'), 'utf8');
+// check resources/adapter/openfin-desktop.js then
+// resources/adapter.asar/openfin-desktop.js
+// for ease of developement
+let jsAdapter = '';
+const searchPaths = ['adapter', 'adapter.asar'];
```
rdepena: I would prefer if this is done via a flag,  we would want to prevent this on non-dev scenarios.

HarsimranSingh: The core has the same behavior without a flag: https://github.com/openfin/runtime/blob/develop/lib/browser/init.js#L115

datamadic: Super edge case... if someone is running in enforced signed assets mode and happens to have that js file sitting there, it will fail first even though a valid asset is present right? Again, I don't think I really care that much...

HarsimranSingh:
🌳 saving for `8.56.25.*`

PR #161, opened 2017-08-31
RUN-3284 Add smallWindow flag and update move/resize functions

HarsimranSingh:
By adding a `smallWindow` flag to allow for very small frameless windows, the move and resize code needs to be updated to preserve bounds-changed and window groups functionality.

runtime: https://github.com/openfin/runtime/pull/608
javascript-adapter: https://github.com/openfin/javascript-adapter/pull/337

### DIFF
--- a/src/browser/animations.js
+++ b/src/browser/animations.js
@@ -246,10 +246,16 @@ function AnimationHandler(desiredInterval) {
                                 });
                             }
                             hwndToId[hwnd] = bw.id;
+                            if (bw.isMaximized()) {
+                                bw.unmaximize();
+                            }
 
                             const { x, y, width: w, height: h } = newBounds;
                             wt.setWindowPos(hwnd, { x, y, w, h, flags });
                         } else {
+                            if (bw.isMaximized()) {
+                                bw.unmaximize();
+                            }
                             bw.setBounds(newBounds);
                         }
                     }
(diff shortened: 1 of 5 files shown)


src/browser/animations.js
```diff
@@ -246,10 +246,16 @@ function AnimationHandler(desiredInterval) {
                                 });
                             }
                             hwndToId[hwnd] = bw.id;
+                            if (bw.isMaximized()) {
```
joneit: you're repeating this 10 times... why not:
```js
bw.demaximize(();
```

And in `Window.create` add:
```js
browserWindow.demaximize = function() {
    if (this.isMaximized()) {
        this.unmaximize();
    }
};
```
If `demaximize` too obscure, how about `unMaximizedIfMaximized`

src/browser/bounds_changed_state_tracker.js
```diff
@@ -130,7 +130,7 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
         var currentBounds = getCurrentBounds();
         var cachedBounds = getCachedBounds();
         var boundsCompare = compareBoundsResult(currentBounds, cachedBounds);
-        var stateMinMax = boundsCompare.state && currentBounds.state !== 'normal'; // maximized or minimized
+        var stateMin = boundsCompare.state && currentBounds.state === 'minimized';
```
HarsimranSingh: Relaxing this condition so it behaves more like v5. We only want to restrict deferred events to minimized windows.

src/browser/animations.js
StevenEBarbaro: Let's refactor this in a second pass

StevenEBarbaro:
👍

src/browser/animations.js
datamadic: is there a second pass for this?

HarsimranSingh:
Will clean up the code in 25

PR #163, opened 2017-08-31
RUN-3299 Handle undefined preload option more gracefully

joneit:
This just avoids logging an error for what it is being validated as a malformed preload option.
Regression [tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59a8814fd2a02971da3a63a1) clean.


HarsimranSingh:
👍

PR #164, opened 2017-09-03
RUN-3300 Cool down code cleanup

joneit:
This mostly: normalizes code that
1. Gets `appIdentity` or `windowIdentity`
2. Microsoft constants

and: Removes `ack` and `nack` from all api handler functions, relying instead on `ElipcStrategy` to:
* normal return from api func _without_ response data: call `ack` with a simple `AckPayload` (just `success: true`)
* normal return from api func _with_ response data: call `ack` with an `AckPayload` with a `data` property
* handles async api funcs (that return a promise).

Integration & Regression [tested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59ac3207d2a02971da3a63a5).

### DIFF
--- a/src/browser/api_protocol/api_handlers/application.js
+++ b/src/browser/api_protocol/api_handlers/application.js
@@ -15,25 +15,25 @@ limitations under the License.
 */
 
 // built-in modules
-let BrowserWindow = require('electron').BrowserWindow;
-let electronApp = require('electron').app;
+const BrowserWindow = require('electron').BrowserWindow;
+const electronApp = require('electron').app;
 
 // npm modules
-let _ = require('underscore');
+const _ = require('underscore');
 
 // local modules
-let Application = require('../../api/application.js').Application;
-let apiProtocolBase = require('./api_protocol_base.js');
-let coreState = require('../../core_state.js');
+const Application = require('../../api/application.js').Application;
+const apiProtocolBase = require('./api_protocol_base.js');
+const coreState = require('../../core_state.js');
 import ofEvents from '../../of_events';
-import {
-    addRemoteSubscription
-} from '../../remote_subscriptions';
+import { addRemoteSubscription } from '../../remote_subscriptions';
 import route from '../../../common/route';
+import { WindowsMessages, SetWindowPosition, SysCommands } from '../../../microsoft';
 
-let successAck = {
-    success: true
-};
+// some local sugar
+const successAck = { success: true };
+const dataAck = data => Object.assign({ data }, successAck);
+const getAppId = message => apiProtocolBase.getTargetApplicationIdentity(message.payload);
 
 module.exports.applicationApiMap = {
     'close-application': closeApplication,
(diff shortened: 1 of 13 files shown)


rdepena:
@joneit some of these changes will not be covered in the integration tests and must be manually validated with the Java docking application.

joneit:
rebased to `8.56.24.30`

joneit:
Regarding making sure all touched methods have tests.... This would be a good opportunity to annotate each API method with its test(s). I'm willing to take this on. We might also consider renaming the tests when possible to "point to" the code (file name, method name).

joneit:
Rebased to `8.56.24.30` and [retested](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59b0cd51d2a02971da3a63d3).

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -78,386 +73,303 @@ function SystemApiHandler() {
...
-        const { payload: { level } } = message;
-        const response = System.setMinLogLevel(level);
-
-        if (didFail(response)) {
-            nack(response);
-
-        } else {
-            ack(_.clone(successAck));
-        }
+    function setMinLogLevel(identity, message) {
```
datamadic: Why did this become a promise?

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -78,386 +73,303 @@ function SystemApiHandler() {
...
-
-        if (didFail(response)) {
-            nack(response);
-
-        } else {
-            const dataAck = _.clone(successAck);
-            dataAck.data = response;
-            ack(dataAck);
-        }
+    function getMinLogLevel(identity, message) {
```
datamadic: Why did this become a promise?

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -78,386 +73,303 @@ function SystemApiHandler() {
...
+        return new Promise((resolve, reject) => {
+            System.deleteCacheOnExit(resolve, reject);
+        });
     }
 
-    function exitDesktop(identity, message, ack) {
-        ack(successAck);
-        System.exit();
+    function exitDesktop(identity, message) {
+        setTimeout(() => System.exit());
```
datamadic: Why is this timed out here? Comment? Test?

src/browser/api_protocol/api_handlers/window.js
```diff
@@ -17,11 +17,10 @@ limitations under the License.
...
 const Window = require('../../api/window').Window;
 const Application = require('../../api/application').Application;
-const _ = require('underscore');
 
-let successAck = {
-    success: true
-};
+// some local sugar
+const getWinId = message => apiProtocolBase.getTargetWindowIdentity(message.payload);
+const getGroupingWinId = message => apiProtocolBase.getGroupingWindowIdentity(message.payload);
```
datamadic: prefer braces

src/browser/api_protocol/api_handlers/window.js
```diff
@@ -17,11 +17,10 @@ limitations under the License.
 const apiProtocolBase = require('./api_protocol_base');
 const Window = require('../../api/window').Window;
 const Application = require('../../api/application').Application;
-const _ = require('underscore');
 
-let successAck = {
-    success: true
-};
+// some local sugar
+const getWinId = message => apiProtocolBase.getTargetWindowIdentity(message.payload);
```
datamadic: `getWinId` might not be the best name as it is actually getting the target window (`getTargetWindowIdentity`) id and these can be different

src/browser/api_protocol/api_handlers/window.js
```diff
@@ -81,474 +80,272 @@ module.exports.init = function() {
...
-    Window.disableFrame(windowIdentity);
-    ack(successAck);
+function disableWindowFrame(identity, message) {
+    Window.disableFrame(getWinId(message));
 }
 
-function windowEmbedded(identity, message, ack) {
-    let payload = message.payload;
-    // Ensure expected shape for identity utility compliance
+function windowEmbedded(identity, message) {
```
datamadic: This needs to be manually tested, is not covered in the test apps. See @nicholasdgoodman

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -40,15 +40,9 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
-                                // older action calls will invoke ack internally, newer ones will return a value
-                                if (result !== undefined) {
-                                    ack(new AckPayload(result));
-                                }
-                            }).catch(err => {
-                                nack(err);
-                            });
+                            .then(() => endpoint.apiFunc(identity, data))
+                            .then(result => { ack(new AckPayload(result)); })
+                            .catch(err => { nack(err); });
```
datamadic: `catch(nack)`

src/browser/api_protocol/api_handlers/application.js
```diff
@@ -15,25 +15,23 @@ limitations under the License.
...
-} from '../../remote_subscriptions';
+import { addRemoteSubscription } from '../../remote_subscriptions';
 import route from '../../../common/route';
+import { WindowsMessages, SetWindowPosition, SysCommands } from '../../../microsoft';
 
-let successAck = {
-    success: true
-};
+// some local sugar
+const getAppId = message => apiProtocolBase.getTargetApplicationIdentity(message.payload);
```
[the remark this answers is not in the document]
joneit: will rename

src/browser/api_protocol/api_handlers/application.js
```diff
@@ -68,377 +66,277 @@ module.exports.init = function() {
...
+    return new Promise((resolve, reject) => {
+        Application.getInfo(getAppId(message), resolve, reject);
+    });
 }
 
-function getParentApplication(identity, message, ack, nack) {
-    const dataAck = _.clone(successAck);
-    const appIdentity = apiProtocolBase.getTargetApplicationIdentity(message.payload);
-    const parentUuid = Application.getParentApplication(appIdentity);
+function getParentApplication(identity, message) {
```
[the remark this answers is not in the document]
joneit: only way to nack

src/browser/api_protocol/api_handlers/system.js
joneit: only way to nack is to return a rejected promise

src/browser/api_protocol/api_handlers/system.js
joneit: will convert to throw

src/browser/api_protocol/api_handlers/system.js
joneit: I'm forcing the actual exit call to come _after_ the ack per original

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
joneit: will change it.
...although technically, that would propagate some unpredictable `error` value (whatever nack returns)... on the other hand the present code doesn't propagate either (returns `undefined`). but it doesn't matter: in any case it's the end of the chain anyway and the promise goes away at that point. but should a future maintainer add another catch, they would need to make sure this one returns its input.

src/browser/api_protocol/api_handlers/window.js
joneit: you mean `x => { return x; }` instead of `x => x` ? 


IMHO that kinda defeats the purpose of
> "An arrow function expression ... has a shorter syntax compared to function expressions ..."
(from [mdn page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Arrow_functions)).

src/browser/api_protocol/api_handlers/window.js
joneit: will rename to `getTargetWinId`

rdepena:
Per conversation this needs to be split up.

PR #168, opened 2017-09-07
RUN-3128 bugfix iframe restore breaking maximize 

datamadic:
Prevent iframes from making windowing api calls on startup

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59b1902cd2a02971da3a63d4)
[added test](https://testing-dashboard.openfin.co/#/app/tests/59b1a9a7d2a02971da3a63d5/edit)

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 /* global fin, window*/
-api
 // These are relative to the preload execution, the root of the proj
 
 // THIS FILE GETS EVALED IN THE RENDERER PROCESS
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -24,6 +24,7 @@ limitations under the License.
     const processVersions = JSON.parse(JSON.stringify(process.versions));
 
     let renderFrameId = global.routingId;
+
     let customData = global.getFrameData(renderFrameId);
     let glbl = global;
 
@@ -357,7 +358,17 @@ limitations under the License.
         //---------------------------------------------------------------
         let winOpts = getWindowOptionsSync();
 
-        showOnReady(glbl, winOpts);
+        // Prevent iframes from attempting to do windowing actions, these will always be handled
+        // by the main window frame.
+        if (window.frameElement) {
+
+            // If you are an iframe, you still need to send out the api ready event so frameConnect
+            // can correctly assign connection priority.
+            electron.remote.getCurrentWebContents(renderFrameId).emit('openfin-api-ready', renderFrameId);
+        } else {
+            showOnReady(glbl, winOpts);
+        }
+
         wireUpMenu(glbl, winOpts);
         wireUpZoomEvents();
         raiseReadyEvents(winOpts);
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -357,7 +357,17 @@ limitations under the License.
...
         let winOpts = getWindowOptionsSync();
 
-        showOnReady(glbl, winOpts);
+        // Prevent iframes from attempting to do windowing actions, these will always be handled
+        // by the main window frame.
+        if (window.frameElement) {
+
+            // If you are an iframe, you still need to send out the api ready event so frameConnect
+            // can correctly assign connection priority.
+            electron.remote.getCurrentWebContents(renderFrameId).emit('openfin-api-ready', renderFrameId);
```
HarsimranSingh: You can move this call out of `showOnReady` and just do it here once

datamadic: 👍 updating and retesting

datamadic:
@sim updated and [tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59b827a03d90dd51ca4f1015), [same as vanilla](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59b82a6f3d90dd51ca4f1019)

HarsimranSingh:
Doesn't the other `openfin-api-ready` event need to be removed from `showOnReady`?

src/renderer/api-decorator.js
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 /* global fin, window*/
-
+api
```
whyn07m3: what is this?

src/renderer/api-decorator.js
```diff
@@ -357,7 +354,16 @@ limitations under the License.
...
-        showOnReady(glbl, winOpts);
+        // Prevent iframes from attempting to do windowing actions, these will always be handled
+        // by the main window frame.
+        if (!window.frameElement) {
+            showOnReady(glbl, winOpts);
+        }
+
+        // The api-ready event allows the webContents to assign api priority. This must happen after
+        // any spin up windowing action or you risk stealing api priority from an already connected frame
+        electron.remote.getCurrentWebContents(renderFrameId).emit('openfin-api-ready', renderFrameId);
```
joneit: are we sure this can be executed before `showOnReady` is "finished" for non-iframe window?

src/renderer/api-decorator.js
datamadic: a mistake :) updating now...

src/renderer/api-decorator.js
datamadic: for iframes the show on ready does not get called, so yeah

joneit: I meant for non-iframe windows, "ready" will be emitted _before_ the "onFinish" callbacks in `showOnReady`.... is that an intentional change? doesn't sound right that we would emit "ready" before min/max/show calls are completed

HarsimranSingh: All the min/max/show functions in `showOnReady` are sync. `showOnReady` can actually be written in a more linear way now that everything is sync (this wasn't the case a while ago).

PR #171, opened 2017-09-13
RUN-3316 Log preload script progress

joneit:
To extract just the preload log items, do something like this:
```bash
grep ' \[PRELOAD\]' /c/Users/Jonathan/AppData/Local/OpenFin/cache/8.56.24.30/debug.log
```
(Note the leading space in the search arg, which gives better filtering of verbose logs, filtering out the `write-to-log` API calls containing [PRELOAD] made by api-decorator which are redundant.)

See sample log output in Jira [RUN-3316](https://appoji.jira.com/browse/RUN-3316).

PR #177, opened 2017-09-18
RUN-3319 Preload scripts not injected on window.open

joneit:
### Issue 1

Default `preloads` was defined (in convert_options::`convertToElectron`) as `[]` (empty array), which was not sufficiently falsy to avoid overwriting underlying parent options.

Code in convert_options was also transforming the legacy form of the `preload` option to modern form. This is a reasonable place for it (despite the outdated name "convertToElectron" which referred to converting v5 options object for use with v6), but is was inadequately implemented as the same logic appeared in at least two other places.

Solution was to:
1. Make the default undefined rather than empty array
2. Remove the "convert" logic repetitions from other places where found
3. Ensure `convertToElectron` is called for all cases
4. Only use `preload` when defined (or convert undefined to empty array before using)
5. Duplicated some of the logic clean-up also found in preload_scripts.ts (see #171 = RUN-3316)

### Issue 2

Main window options are correctly converted when they are first used. However, unlike with `win._options`, the converted main window options are _not_ saved in `app._options` (`eOpts`, defined [here](https://github.com/HadoukenIO/core/blob/runtime-v8.56.24.39/src/browser/api/application.js#L987), is not being saved). So this is the original (unmutated) object. The problem is it is also the source of core_state::`getMainWindowOptions()`. I think a better source might be the actual main window object's `_options` object (if there is such a thing) — because it _is_ converted.

The problem with this situation is that the code was doing [this](https://github.com/HadoukenIO/core/pull/177/files#diff-d5752a5b2c3b7ce27c77d78f1b9278a6L398):
```js
_options = _.extend(_.clone(baseOpts), convertOptions.convertToElectron(opts));
```
The above draws from the unconverted main window options. This has the unfortunate effect that when the parent option is _not_ being overridden you can unexpectedly end up with an unconverted value. Possible reason this hasn't been (much of) an issue before is that for the most part "conversion" here means copying values to different option names. Only the `preload` option really does critical transformations on its value (which is why I didn't want this code here in the first place, but now it seems reasonable to put it here).

Solution: Although I think we should fix `GetMainWindowOptions` (to use the window's `_options` object, if available, rather than the app's). For now I was able to make this work changing the above to [this](https://github.com/HadoukenIO/core/pull/177/files#diff-d5752a5b2c3b7ce27c77d78f1b9278a6R398):
```js
_options = convertOptions.convertToElectron(Object.assign({}, baseOpts, opts));
```
Ignore the introduction of `assign`, which is not the point; the important part is this: Rather than converting (just) the child window options and superimposing them over the parent options, this new version first does the superimposing of the raw options; then does the converting (including converting the main window's options).

But I also had to make sure to call `convertToElectron` again for the `app._options` usage [here](https://github.com/HadoukenIO/core/pull/177/files#diff-a08b93ff0f46d39099ebdc2bb86958a1R448). So really this is getting messier and it would be better to get rid of `app._options`.

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -265,14 +265,18 @@ module.exports = {
         }
 
         const preload = options.preload;
-        if (!preload) {
-            // for all falsy values
-            newOptions.preload = [];
-        } else if (typeof preload === 'string') {
-            // backward compatibility
-            newOptions.preload = [{ url: preload }];
-        } else {
-            newOptions.preload = preload;
+        if (preload) {
+            if (typeof preload === 'string') {
+                // convert legacy `preload` option into modern `preload` option
+                newOptions.preload = [{ url: preload }];
+            } else if (
+                Array.isArray(preload) &&
+                preload.every(eachPreload => typeof eachPreload === 'object' && typeof eachPreload.url === 'string')
+            ) {
+                newOptions.preload = preload;
+            } else {
+                log.writeToLog('warning', 'Expected `preload` option to be a string primitive OR an array of objects with `url` string properties.');
+            }
         }
 
         if (options.customRequestHeaders !== undefined) {
--- a/src/browser/preload_scripts.ts
+++ b/src/browser/preload_scripts.ts
@@ -54,7 +54,7 @@ type LoadResponses = boolean[];
 /** Returns a `Promise` once all preload scripts have been fetched and loaded or have failed to fetch or load.
  *
  * @param {object} identity
- * @param {string|object[]} preloadOption
+ * @param {PreloadInstance[]} preloadOption
  * @param {function} [proceed] - If supplied, both `catch` and `then` are called on it.
  *
  * If you don't supply `proceed`, call both `catch` and `then` on your own to proceed as appropriate.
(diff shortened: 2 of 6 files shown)


src/renderer/api-decorator.js
```diff
@@ -426,11 +426,14 @@ limitations under the License.
...
         // Reset state machine values that are set through synchronous handshake between native WebContent lifecycle observers and JS
         options.openfin = true;
-        options.preload = 'preload' in options ? options.preload : winOpts.preload;
+
+        // Force window to be a child of its parent application.
         options.uuid = winOpts.uuid;
+
         // Apply parent window background color to child window when child
         // window background color is unspecified.
         options.backgroundColor = options.backgroundColor || winOpts.backgroundColor;
```
joneit: Question about this existing code:
Why is this needed here? Is `Window.create` being called? (How, btw?) If so, it inherits options from main window...

joneit: Question about this existing code:
Why is this needed here? Is `Window.create` being called? (How, btw?) If it is, note that it already inherits from main window options (see [above](https://github.com/HadoukenIO/core/pull/177/files#diff-d5752a5b2c3b7ce27c77d78f1b9278a6L398)).

src/browser/preload_scripts.ts
```diff
@@ -72,60 +74,46 @@ export function fetchAndLoadPreloadScripts(
...
-                return Promise.resolve();
-            } else {
-                // not previously downloaded *OR* previous downloaded failed
-                return fetch(identity, preload).then(load);
-            }
-        });
+): Promise<LoadResponses> {
+    let result: Promise<LoadResponses>;
+
+    const loadedScripts: Promise<undefined>[] = preloadOption.map((preload: PreloadInstance) => {
```
rdepena: @joneit this will be removed in the updated call right?

joneit: yup, a lot of this is out of date. i'm merging in prior commits. it will change again in subsequent commits.

joneit:
Rebased to 1bc274e (current develop, which is 3 commits ahead of `8.56.25.7`).
Testing in progress...

src/browser/convert_options.js
```diff
@@ -265,14 +265,18 @@ module.exports = {
...
+            if (typeof preload === 'string') {
+                // convert legacy `preload` option into modern `preload` option
+                newOptions.preload = [{ url: preload }];
+            } else if (
+                Array.isArray(preload) &&
+                preload.every(eachPreload => typeof eachPreload === 'object' && typeof eachPreload.url === 'string')
+            ) {
+                newOptions.preload = preload;
+            } else {
+                log.writeToLog('warning', 'Expected `preload` option to be a string primitive OR an array of objects with `url` string properties.');
```
rdepena: Lets not fix this right now because its out of scope but we should error out if the preload script format is wrong to the point we cannot take any actions, never fail silently.

rdepena:
👍

src/browser/convert_options.js
datamadic: https://appoji.jira.com/browse/RUN-3371

src/browser/preload_scripts.ts
```diff
@@ -76,38 +76,25 @@ export function fetchAndLoadPreloadScripts(
...
-            } else {
-                // not previously downloaded *OR* previous downloaded failed
-                return fetchToCache(identity, preload).then(loadFromCache);
-            }
-        });
+    const loadedScripts: Promise<undefined>[] = preloadOption.map((preload: PreloadInstance) => {
+        // following if clause avoids re-fetch for remote resources already in memory
+        // todo: following if clause slated for removal (RUN-3227, blocked by RUN-3162), i.e., return always
+        if (
+            !REGEX_FILE_SCHEME.test(preload.url) && // not a local file AND...
```
datamadic: what if we just let it try to fetch an handle the fail?

joneit: not to worry :) these lines are removed entirely by my next PR #181 (RUN-3321 electron.net).

PR #178, opened 2017-09-19
RUN-3335 Normalize Microsoft constants

joneit:
Also collapsed some whitespace.
[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59c16f307cb79f732d10c36a)

### DIFF
--- a/src/browser/api_protocol/api_handlers/application.js
+++ b/src/browser/api_protocol/api_handlers/application.js
@@ -26,10 +26,9 @@ let Application = require('../../api/application.js').Application;
 let apiProtocolBase = require('./api_protocol_base.js');
 let coreState = require('../../core_state.js');
 import ofEvents from '../../of_events';
-import {
-    addRemoteSubscription
-} from '../../remote_subscriptions';
+import { addRemoteSubscription } from '../../remote_subscriptions';
 import route from '../../../common/route';
+import { WindowsMessages, SetWindowPosition, SysCommands } from '../../../common/microsoft';
 
 let successAck = {
     success: true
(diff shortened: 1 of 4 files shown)


HarsimranSingh:
👍

joneit:
Rebased to `8.56.25.3`. (Will rerun tests closer to merge time in case another rebase is required.)

src/browser/api_protocol/api_handlers/application.js
```diff
@@ -350,71 +340,58 @@ function deregisterExternalWindow(identity, message, ack) {
...
             ofEvents.emit(route.externalWindow('close', uuid, name));
             break;
-        case 7:
-            // WM_SETFOCUS
+        case WindowsMessages.WM_SETFOCUS:
             ofEvents.emit(route.externalWindow('focus', uuid, name));
             break;
-        case 8:
-            // WM__KILLFOCUS
+        case WindowsMessages.WM__KILLFOCUS:
```
datamadic: This is undefined... `WM__KILLFOCUS` vs `WM_KILLFOCUS`. How can we confirm these?

joneit: we could do WindowMessages.get('KILLFOCUS') which could log an error (or worse) if KILLFOCUS is unknown, but would be unusual coding and probably overkill.

joneit: I searched for `__` inside a word and the above was the only hit:
```bash
egrep -r '\w+__\w+' src
```

wenjunche:
👍

PR #179, opened 2017-09-20
RUN-3049 Extend drag area for frameless window

TomaszMa:
The default drag are really small I extended it to 4 pixels and 6 pixels on corners.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -164,8 +164,8 @@ let optionSetters = {
             // reapply resize region
             applyAdditionalOptionsToWindowOnVisible(browserWin, () => {
                 let resizeRegion = getOptFromBrowserWin('resizeRegion', browserWin, {
-                    size: 2,
-                    bottomRightCorner: 4
+                    size: 4,
+                    bottomRightCorner: 6
                 });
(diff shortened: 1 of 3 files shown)


joneit:
It's about time. :) Framed windows seem to have a **7-pixel** edge grab region. I don't know if 7 is a Windows standard, but it _is_ very comfortable. (The fact that the content area is rendered _inside_ the region instead of superimposed on top of it notwithstanding.) Should we consider 7 for frameless as well?

rdepena:
👍

HarsimranSingh:
If this change is happening, we need to update the docs

TomaszMa:
Doc update: https://github.com/openfin/javascript-adapter/pull/342

datamadic:
In addition, ideally this is accompanied with a test in our automated testing suite

TomaszMa:
@datamadic @whyn07m3 Agree with you guys, but do we have place where I could add these two constants, or there in no source file for this purpose, and I need to create it.

whyn07m3:
@TomaszMa IMO, the "closest", but not the best, location where you could put these constants would be `src/shapes.ts`. Maybe if we have more of such constants throughout the project later on, we can create a separate file.

PR #180, opened 2017-09-21
RUN-3038 Load plugin modules

VanessaSergeon:
Given a plugin entry in the app.json, the runtime will need to determine the folder with preload plugins. 

[Wiki to be moved to Core repo](https://github.com/openfin/Internal-Wiki/wiki/RVM-to-Runtime-Plugin-Query)

[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59cd51e27cb79f732d10c3f3)  9/12 failing tests pass in isolation.

[Manual tests PR](https://github.com/openfin/test_apps/pull/8)

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -766,9 +766,10 @@ Window.create = function(id, opts) {
 
     // Set preload scripts' final loading states
     winObj.preloadState = _options.preload.map(preload => {
+        const identitier = preload.url ? preload.url : preload.name;
         return {
-            url: preload.url,
-            state: getPreloadScriptState(preload.url)
+            url: identitier,
+            state: getPreloadScriptState(identitier)
         };
     });
 
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -38,23 +38,17 @@ let clipBounds = require('../clip_bounds.js').default;
 let convertOptions = require('../convert_options.js');
 let coreState = require('../core_state.js');
 let ExternalWindowEventAdapter = require('../external_window_event_adapter.js');
-import {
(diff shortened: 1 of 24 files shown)


src/browser/api/window.js
```diff
@@ -2165,4 +2165,8 @@ function getElectronBrowserWindow(identity, errDesc) {
     return browserWindow;
 }
 
+function getIdentifier(preload) {
+    return preload.url ? preload.url : `${preload.name}-${preload.version}`;
```
rdepena: @VanessaSergeon why is getIdentifier also defined here ?

VanessaSergeon: Oh thanks for catching that duplication. Will import from preload_scripts.ts instead.

rdepena:
👍

src/browser/preload_scripts.ts
```diff
@@ -136,37 +152,66 @@ function fetchToCache(identity: Identity, preloadScript: PreloadInstance): Promi
...
+            action: 'query-plugin',
+            name: preloadPlugin.name,
+            version: preloadPlugin.version,
+            optional: preloadPlugin.optional,
+            sourceUrl
+        };
+
+        rvmMessageBus.publish(msg, (resp) => {
+            if (resp.payload.hasOwnProperty('path') && resp.payload.action === 'query-plugin') {
+                const scriptPath = `${resp.payload.path}\\${resp.payload.target}`;
```
datamadic: do we need to check that the `resp.payload.path` does not already end in a `\`? In general we should probably use the node `path` utility for stuff like this

PR #186, opened 2017-09-25
RUN-3366 writeToLog use toString when avail

joneit:
Please see Jira [RUN-3366](https://appoji.jira.com/browse/RUN-3366) for explanation.

**[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59c96bde7cb79f732d10c3ab).**

This is another casual general improvement PR and can slip to next sprint considering this sprint has so much to merge already.

### DIFF
--- a/src/browser/log.ts
+++ b/src/browser/log.ts
@@ -33,25 +33,27 @@ export const logLevelMappings = new Map<any, any>([
  * Parses log messages and uses Electron's APIs to log them to console
  */
 export function writeToLog(level: any, message: any, debug?: boolean): any {
+    const isObj = typeof message === 'object';
     let parsedMessage: string;
 
     // Parse log message
     try {
 
-        if (typeof message === 'object') {
-            if (message instanceof Error) {
+        if (isObj && message instanceof Error) {
 
-                // Properly stringify error objects
-                parsedMessage = JSON.stringify(errorToPOJO(message));
-            } else {
+            // Properly stringify error objects (i.e., stack and message properties only)
+            parsedMessage = JSON.stringify(errorToPOJO(message));
 
-                // Stringify plain objects
-                parsedMessage = JSON.stringify(message);
-            }
+        } else if (isObj && (message === null || message.toString === Object.prototype.toString)) {
+
+            // Don't use Object's toString which just returns "[object Object]"
+            parsedMessage = JSON.stringify(message);
+
         } else {
 
-            // Convert non-object messages to a string
-            parsedMessage = message.toString();
+            // Use object's custom toString function OR convert primitive values to string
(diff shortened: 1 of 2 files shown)


whyn07m3:
The following is properly being stringified in the `writeToLog` function:

```
log.writeToLog('info', { a: 'lol' });
log.writeToLog('info', ['l', 0, 'l']);
```

What is this PR trying to fix?

wenjunche:
can it skip errorToPOJO and stringify based on verboseLogLevel ?

joneit:
@whyn07m3 please see Jira for answer to your question. when an object has a custom toString, we should use it

joneit:
@wenjunche please elucidate... re `errorToPOJO` we should always do that because some error objects resists stringification otherwise.

whyn07m3:
I haven't come across this use case (Object's custom `.toString`) in our project... Can you please give an example of a real life scenario?..

joneit:
@whyn07m3 `Date` is an example of an object that has its own `toString` which produces something useful (`Sun Oct 01 2017 12:11:11 GMT-0400 (EDT)`). There is no need to stringify.

`JSON.stringify` is not meant for general serialization; it is meant to produce transmittable JSON. Therefore, it is not a completely reliable serializer.

In particular, it will only output representations of the object's own, enumerable properties, ignoring functions and `undefined` values.

Also, objects that implement `toJSON` specifically for data output. `Date` is one such. For example, `Date.toJSON()` produces `"2017-10-01T16:11:11.799Z"` (including the quotes) which is a precise string that all database engines understand (it's also collatable). This value may or may not be suitable for logging. (I didn't go there in this PR.)

Classic use of `toString` for objects is to produce a string representation of the object. `Object.toString` is pretty much useless and _is meant to be overridden._ We've been using `JSON.stringify` because we don't want to log `[Object object]` (produced by `Object.toString`). Fair enough. But when `toString` is overridden, we should use the override. 

This PR simply checks if the object implements `toString` and uses it when it does, before resorting to `JSON.stringify`.

whyn07m3:
@joneit Thank you for providing the use case and explaining! 👍

joneit:
Now prepends class name (when known*) to object stringification:
```
[10/03/17 10:46:36]-[INFO:atom_api_app.cc(1034)] 10/03 10:46:36.118 - MyClass: {"a":17}
```
[**Retested**](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59d3a2d07cb79f732d10c428).
\___________
\* when object has a `constructor` property and `constructor.name` is neither `'Object'` (plain object; not a class) nor `'Function'` (is a class but anonymous).

joneit:
@HarsimranSingh Added unit tests.

PR #191, opened 2017-10-02
RUN-3127 Change default backgroundColor to #FFF

TomaszMa:
Currently default backgroundColor is set to #000, so when backgroundColor in other PR will be fixed the background color of js apps will be black. Currently background color of js apps are white because "backgroundColor" of js api doesn't work correctly. First to fix "backgroundColor" functionality I need to change default background color.

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -104,7 +104,7 @@ function five0BaseOptions() {
         'url': 'about:blank',
         'uuid': '',
         'waitForPageLoad': true,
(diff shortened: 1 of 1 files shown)


HarsimranSingh:
This needs a docs change https://github.com/openfin/javascript-adapter/blob/develop/src/window.js#L336

TomaszMa:
PR for doc update: https://github.com/openfin/javascript-adapter/pull/343

PR #208, opened 2017-10-20
RUN-3390 (phase 2) separate plugins from preload scripts

whyn07m3:
This is **Phase 2** of _plugin-preload separation/refactor/cleanup/fix_.

All plugin/preload tests will be written when all phases are completed.

This 2nd phase:
• separates plugin logic from preload scripts logic in places that deal with plugin retrieval

✅ Test results:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59e9fea1af26a25be2ffc907)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59e9feecaf26a25be2ffc908)

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -41,6 +41,7 @@ const ProcessTracker = require('../process_tracker.js');
 import route from '../../common/route';
 import { fetchAndLoadPreloadScripts, getIdentifier } from '../preload_scripts';
 import { FrameInfo } from './frame';
+import * as plugins from '../plugins';
 
 const defaultProc = {
     getCpuUsage: function() {
@@ -698,6 +699,10 @@ exports.System = {
         } else {
             fetchAndLoadPreloadScripts(identity, preloadOption, cb);
         }
+    },
+
+    getPluginModules: function(cb) {
+        return plugins.getModules().then(cb);
     },
 
     // identitier is preload script url or plugin name
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -701,8 +701,8 @@ exports.System = {
         }
     },
 
-    getPluginModules: function(cb) {
-        return plugins.getModules().then(cb);
+    getPluginModules: function() {
+        return plugins.getModules();
     },
 
     // identitier is preload script url or plugin name
--- a/src/browser/core_state.ts
+++ b/src/browser/core_state.ts
@@ -209,13 +209,13 @@ export function getUuidBySourceUrl(sourceUrl: string): string|boolean {
     return app && app.appObj && app.appObj.uuid;
 }
 
-export function getConfigUrlByUuid(uuid: string): string|boolean {
+export function getConfigUrlByUuid(uuid: string): string {
     const app = getAppAncestor(uuid);
     if  (app && app._configUrl) {
         return app._configUrl;
     } else {
         const externalApp = getExternalAncestor(uuid);
-        return externalApp && externalApp.configUrl;
+        return (externalApp && externalApp.configUrl) || '';
     }
 }
 
--- a/src/browser/plugins.ts
+++ b/src/browser/plugins.ts
@@ -38,17 +38,15 @@ export async function getModules(): Promise<PluginWithContent[]> {
  */
 async function getModule(sourceUrl: string, plugin: Plugin): Promise<PluginWithContent> {
     const id = `${plugin.name}: ${plugin.version}`;
+    let pluginPath;
 
-    // Plugin path is already available
     if (pluginPaths.has(id)) {
-        const pluginPath = pluginPaths.get(id);
-        return await addContent(plugin, pluginPath);
+        pluginPath = pluginPaths.get(id);
+    } else {
+        const {payload} = await rvmMessageBus.getPluginInfo(sourceUrl, plugin);
+        pluginPath = !payload.error ? path.join(payload.path, payload.target) : '';
+        pluginPaths.set(id, pluginPath);
     }
-
-    const {payload} = await rvmMessageBus.getPluginInfo(sourceUrl, plugin);
-    const pluginPath = !payload.error ? path.join(payload.path, payload.target) : '';
-
-    pluginPaths.set(id, pluginPath);
 
     return await addContent(plugin, pluginPath);
 }
--- a/src/browser/plugins.ts
+++ b/src/browser/plugins.ts
@@ -1,0 +1,73 @@
+/*
+Copyright 2017 OpenFin Inc.
+
+Licensed under OpenFin Commercial License you may not use this file except in compliance with your Commercial License.
+Please contact OpenFin Inc. at sales@openfin.co to obtain a Commercial License.
+*/
+
+import * as path from 'path';
+import { getStartManifest, StartManifest } from './core_state';
+import { Plugin } from '../shapes';
+import { readFile } from 'fs';
+import { rvmMessageBus } from './rvm/rvm_message_bus';
+
+interface PluginWithContent extends Plugin {
+    _content: string;
+}
+
+/**
+ * A map of already-retrieved plugin module paths.
+ *
+ * Examples:
+ * 'pluginName: 1.0.0' -> 'path\to\plugin\injectable\script.js'
+ * 'pluginName2: 0.0.1' -> '' // plugin/version doesn't exist
+ */
+const pluginPaths: Map<string, string> = new Map();
+
+/**
+ * Gets all plugins defined in app's manifest
+ */
+export async function getModules(): Promise<PluginWithContent[]> {
+    const {url, data: {plugin: plugins = []}} = <StartManifest>getStartManifest();
+    const promises = plugins.map((plugin: Plugin) => getModule(url, plugin));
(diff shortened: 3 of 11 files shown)


whyn07m3:
✅ Re-tested after the rebase:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f35fa0af26a25be2ffc985)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f35fedaf26a25be2ffc986)

src/browser/api/system.js
```diff
@@ -700,6 +701,10 @@ exports.System = {
         }
     },
 
+    getPluginModules: function(cb) {
+        return plugins.getModules().then(cb);
```
datamadic: Does this need a `.catch()` to prevent the case where an error keeps the callback / ack waiting for ever?

src/browser/plugins.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Examples:
+ * 'pluginName: 1.0.0' -> 'path\to\plugin\injectable\script.js'
+ * 'pluginName2: 0.0.1' -> '' // plugin/version doesn't exist
+ */
+const pluginPaths: Map<string, string> = new Map();
+
+/**
+ * Gets all plugins defined in app's manifest
+ */
+export async function getModules(): Promise<PluginWithContent[]> {
```
datamadic: For all of these `async` calls, if the only place they are ever called is form the api_handler then perhaps not handling the fails is OK as that is explicitly try caught, but we need to be sure that these cannot blow up browser side outside a try/catch. This may already be the case.

src/browser/plugins.ts
```diff
@@ -0,0 +1,73 @@
...
+    const {url, data: {plugin: plugins = []}} = <StartManifest>getStartManifest();
+    const promises = plugins.map((plugin: Plugin) => getModule(url, plugin));
+    return await Promise.all(promises);
+}
+
+/**
+ * Gets a single plugin module
+ */
+async function getModule(sourceUrl: string, plugin: Plugin): Promise<PluginWithContent> {
```
datamadic: :neckbeard: you can avoid an early return here by `if/else` setting the plugin path

src/browser/plugins.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    pluginPaths.set(id, pluginPath);
+
+    return await addContent(plugin, pluginPath);
+}
+
+/**
+ * Reads and adds module content to plugin
+ */
+function addContent(plugin: Plugin, pluginPath: string): Promise<PluginWithContent> {
```
datamadic: is this async? It can still be `awaited` without being a promise

src/browser/preload_scripts.ts
```diff
@@ -102,11 +93,7 @@ export function fetchAndLoadPreloadScripts(
             return Promise.resolve(true);
         } else {
             // not previously downloaded *OR* previous downloaded failed
-            if (isPreloadScript(preload)) {
-                return fetchToCache(identity, preload).then(loadFromCache);
-            } else {
-                return fetchPlugin(identity, preload).then(loadFromCache);
-            }
+            return fetchToCache(identity, preload).then(loadFromCache);
```
datamadic: Is the fail case something we may need to handle?

src/browser/preload_scripts.ts
```diff
@@ -152,41 +139,14 @@ function fetchToCache(identity: Identity, preloadScript: PreloadScript): Promise
...
-
 // resolves to type `PreloadLoaded` on success
 // resolves to `undefined` when above fetch failed or when successfully fetched asset fails to load from Chromium cache
 function loadFromCache(opts: FetchResponse): Promise<boolean> {
     return new Promise((resolve: Resolver, reject: Rejector) => {
         if (!opts || !opts.scriptPath) {
             resolve(false); // got fetchError above OR no error but no scriptPath either; in any case don't attempt to load
         } else {
-            const preload = opts.preloadScript ? opts.preloadScript : opts.preloadPlugin;
+            const preload = opts.preloadScript;
```
datamadic: `{ destructure }`?

src/browser/rvm/rvm_message_bus.ts
```diff
@@ -352,6 +356,26 @@ export class RVMMessageBus extends EventEmitter  {
             }
         }
     }
+
+    /**
+     * Retrieves information about a plugin
+     */
+    public getPluginInfo(manifestUrl: string, opts: Plugin): Promise<PluginQueryResponse> {
```
datamadic: goes this `getPluginInfo` or `requestPluginInfo`?

src/renderer/api-decorator.js
```diff
@@ -569,51 +569,43 @@ limitations under the License.
...
-
-        try {
-            plugins = syncApiCall('get-selected-preload-scripts', pluginOption);
-        } catch (error) {
-            return syncApiCall('write-to-log', { level: 'error', message: logBase + error });
-        }
 
         plugins.forEach((plugin) => {
-            const { name, version, content } = plugin;
+            const { name, version, _content } = plugin;
```
datamadic: why the `_`?

src/browser/api/system.js
whyn07m3: None of the functions it goes through fail, so should be okay

src/browser/plugins.ts
whyn07m3: Should be fine / okay

src/browser/plugins.ts
whyn07m3: Early `return` in this function prevents the following RVM request

src/browser/plugins.ts
whyn07m3: This one returns an explicit promise that handles both success/failure cases

src/browser/preload_scripts.ts
whyn07m3: This one was a quick edit for other changes in this PR without changing internal functions. We can do it, but in a separate PR though

src/browser/preload_scripts.ts
whyn07m3: Tried to minimize changes to this file as this change was a quick one for other main changes in this PR.

src/browser/api/system.js
datamadic: :smile: none are _designed_ to fail

src/browser/plugins.ts
datamadic: :+1:

src/browser/plugins.ts
datamadic: I know, but if you put it in the else it wont hit it either

src/browser/rvm/rvm_message_bus.ts
whyn07m3: I can certainly change to `requestPluginInfo`, no problem... what's the main reason for change?

src/browser/plugins.ts
datamadic: :+1: I see it now, morning eyes

src/renderer/api-decorator.js
whyn07m3: This was done to separate plugin options. So `_content` will have the content of plugin modules, the rest of the options are original, whatever was set by the user in the manifest. The original options are needed here for logging purposes

datamadic: `options` here are the value `plugin` on line 590?

whyn07m3: Yeah so `plugin` here is the object that has original plugin options from the manifest + `_content` which has text of the module to be eval'ed

whyn07m3:
REVIEW ME PLEASE 👀

✅ Re-tested after the applying comments' suggestions:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f75471af26a25be2ffc98f)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f754b9af26a25be2ffc990)

wenjunche:
Missing some code for updating states for iframes, but the professor said it is for phase 3+
👍

src/browser/core_state.ts
```diff
@@ -209,13 +209,13 @@ export function getUuidBySourceUrl(sourceUrl: string): string|boolean {
     return app && app.appObj && app.appObj.uuid;
 }
 
-export function getConfigUrlByUuid(uuid: string): string|boolean {
+export function getConfigUrlByUuid(uuid: string): string {
```
rdepena: a Nick somewhere got his wings.

src/renderer/api-decorator.js
rdepena: this could be clarified a bit, not sure I could tell by just reading it.

whyn07m3:
@wenjunche RUN-3513

src/browser/rvm/rvm_message_bus.ts
datamadic: just semantics, it does not return or get anything, just posts a message. Presumably the RVM will answer with what we need on a message we are subscribed to somewhere else

PR #213, opened 2017-10-23
RUN-3471: client cert select dialog not working

wenjunche:
Test results:
[W7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59ee2222af26a25be2ffc912)
[W10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59ee2267af26a25be2ffc913)

[Runtime PR](https://github.com/openfin/runtime/pull/669)

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,8 +99,10 @@ app.on('select-client-certificate', function(event, webContents, url, list, call
         skipTaskbar: true,
         resizable: false,
         alwaysOnTop: true,
-        nodeIntegration: true,
-        openfinIntegration: false
+        webPreferences: {
+            nodeIntegration: true,
+            openfinIntegration: false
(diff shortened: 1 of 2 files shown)


wenjunche:
To test, it needs a https server that requires client certificates and the same client certificates need to be imported into Windows cert store, so at least need a new manual test server instead of hosting it on S3

weiteho:
1. see below screenshot, popup shows up at (0,0), which is not the best place. It should be either center of the screen, or center and over requesting window

2. click "cancel" or "X" button has no effect.





image

wenjunche:
The dialog was originally created with RUN-1767.  I created a new Jira, RUN-3514, to make the dialog great again.

PR #215, opened 2017-10-24
RUN-3328 (phase 1) multi-runtime mesh system calls

tgoc99:
First phase includes system calls that need to be aggregated together from other runtimes:

- getAllApplications
- getAllExternalApplications
- getAllWindows
- getProcessList

[JS-Adapter Tests](https://github.com/HadoukenIO/js-adapter/pull/50)
[Middleware wiki:](https://github.com/HadoukenIO/core/wiki/API-Pipeline-Middleware) See locals addition at the bottom


src/browser/api_protocol/api_handlers/system.js
```diff
@@ -224,14 +219,44 @@ function SystemApiHandler() {
...
+        const { locals } = message;
+        let dataAck = _.clone(successAck);
+        dataAck.data = System.getAllExternalApplications();
+
+        if (locals && locals.aggregate) {
+            const { aggregate } = locals;
+            const version = System.getVersion();
+            const portInfo = coreState.getSocketServerState();
+            const { port, securityRealm } = portInfo;
+            const currentApplication = securityRealm ? `${version}/${port}/${securityRealm}` : `${version}/${port}/`;
```
rdepena: :neckbeard: you can use the connection_manager.ts->getMeshUuid function here.

src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -148,11 +156,39 @@ function ferryActionMiddleware(msg: MessagePackage, next: () => void) {
...
+    const { identity, data, ack, nack } = msg;
+    const action = data && data.action;
+    const isAggregateAction = apiMessagesToAggregate[action];
+    //runtimeUuid as part of the identity means the request originated from a different runtime. We do not want to handle it.
+    const isLocalAction = !identity.runtimeUuid;
+
+    try {
+        if (connectionManager.connections.length && isAggregateAction && isLocalAction) {
+            Promise.all(connectionManager.connections.map(runtime => runtime.fin.System.executeOnRemote(identity, data)))
+            .then(data => {
```
datamadic: :neckbeard: shadowing `data`

src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -148,11 +156,39 @@ function ferryActionMiddleware(msg: MessagePackage, next: () => void) {
...
+                const externalRuntimeData = data.reduce((result, runtime) => [...result, ...runtime.data], []);
```
datamadic: My only concern here is that, though right now all the calls return arrays, using `reduce` here over something like `for ... of` means that if ever want to expand this to a call that returns an `object` or `map` in the future this will need to be redone. This does not really bother me too much, but perhaps reduce is less flexible than other.

src/browser/api_protocol/transport_strategy/base_handler.ts
```diff
@@ -17,8 +17,12 @@ const system = require('../../api/system').System;
 export default class RequestHandler<T> {
     private handlers: Array<any> = [];
 
-    private mkNext(fn: any, msg: T) {
-        return () => {
+    private mkNext(fn: any, msg: any) {
```
datamadic: Why `msg` is not longer templated?

tgoc99:
Made requested changes, re-tested on the hadouken-js-adapter tests.

src/browser/api_protocol/api_handlers/mesh_middleware.ts
tgoc99: OK - will keep concept in mind going fwd.

PR #220, opened 2017-10-31
RUN-3472 fixes fetching non-existent preload scripts

whyn07m3:
ℹ️ This PR fixes fetching non-existent preload scripts, which in turn fixes their states

➕ New test: [window.addEventListener (preload-state-changing) (load-failed)](https://testing-dashboard.openfin.co/#/app/tests/59ee481baf26a25be2ffc920/edit)

✅ Test results:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f89658af26a25be2ffc9aa)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f896c3af26a25be2ffc9ab)

PR #221, opened 2017-10-31
RUN-3321 RUN-3434 move to net module

datamadic:
@HarsimranSingh @whyn07m3 

This swaps out our custom, native backed `resourceFetcher` with the standard electron [net](https://electron.atom.io/docs/api/net/) module. We still will write the response to disk (we can eval later if keeping in memory is appropriate for all cases, I wish it was, as it is already on disc per Chrome cache rules). 

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/59f8bcf6af26a25be2ffc9b9)

### DIFF
--- a/src/browser/cached_resource_fetcher.ts
+++ b/src/browser/cached_resource_fetcher.ts
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {app, resourceFetcher, net} from 'electron'; // Electron app
+import {app, net} from 'electron'; // Electron app
 import {stat, mkdir, writeFileSync, createWriteStream} from 'fs';
 import {join, parse} from 'path';
 import {parse as parseUrl} from 'url';
@@ -133,7 +133,7 @@ function generateHash(str: string): string {
 /**
  * Downloads the file from given url using Resource Fetcher and saves it into specified path
  */
-function download(fileUrl: string, filePath: string): Promise<any> {
+function download(fileUrl: string, filePath: string): Promise<void> {
     return new Promise((resolve, reject) => {
         const request = net.request(fileUrl);
         const binaryWriteStream = createWriteStream(filePath, {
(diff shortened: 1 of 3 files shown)


src/browser/cached_resource_fetcher.ts
```diff
@@ -117,17 +125,28 @@ function generateHash(str: string): string {
...
+                    resolve();
+                });
+                binaryWriteStream.once('error', (err: Error) => {
+                    reject(err);
+                });
+                binaryWriteStream.end();
+            });
+        });
+        request.end();
     });
```
wenjunche: no request.on('error', ...) and response.on('error', ...) ?

src/browser/cached_resource_fetcher.ts
```diff
@@ -54,37 +57,42 @@ export function cachedFetch(appUuid: string, fileUrl: string, callback: (error:
...
+    let err: Error;
+
+    try {
+        await prepDownloadLocation(appCacheDir, filePath);
+        await download(fileUrl, filePath);
+    } catch (e) {
+        err = e;
+    }
+
+    callback(err, filePath);
```
whyn07m3: In addition to callbacks, may/should/will this async function return value(s). Maybe it's not needed, just asking...

src/browser/cached_resource_fetcher.ts
```diff
@@ -13,8 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {app, resourceFetcher} from 'electron'; // Electron app
-import {stat, mkdir} from 'fs';
+import {app, net} from 'electron'; // Electron app
+import {stat, mkdir, writeFileSync, createWriteStream} from 'fs';
```
HarsimranSingh: Remove writeFileSync if it isn't needed

HarsimranSingh:
You're gonna need to account for this since it's getting merged in first

https://github.com/HadoukenIO/core/pull/220

rdepena:
@datamadic How will this treat 302 response codes, do we need to handle asset acquisition or will the module follow redirects ?

datamadic:
@rdepena looks like redirect works just fine
https://testing-dashboard.openfin.co/#/app/tests/59fc9771af26a25be2ffca2b/edit

PR #222, opened 2017-10-31
RUN-3481 add opacity to notifications

tgoc99:
Add opacity option to notifications.

[Related JS-Adapter PR](https://github.com/openfin/javascript-adapter/pull/352)

### DIFF
--- a/src/browser/api/notifications/observable_sequences.ts
+++ b/src/browser/api/notifications/observable_sequences.ts
@@ -42,12 +42,12 @@ export const noteStack = Rx.Observable.merge(
             opacity: x.options.opacity
         };
     }),
-    removes.map((x: any) => {
+    removes.map((x: Identity) => {
         return {
             create: 0,
             name: x.name,
             uuid: x.uuid,
-            opacity: x.opacity
+            opacity: null
         };
     }))
     .distinctUntilChanged()
--- a/src/browser/api/notifications/observable_sequences.ts
+++ b/src/browser/api/notifications/observable_sequences.ts
@@ -39,7 +39,7 @@ export const noteStack = Rx.Observable.merge(
             create: 1,
             name: x.identity.name,
             uuid: x.identity.uuid,
-            opacity: x.options.opacity
+            opacity: x.options.finalOpacity
         };
     }),
     removes.map((x: Identity) => {
--- a/src/browser/api/notifications/observable_sequences.ts
+++ b/src/browser/api/notifications/observable_sequences.ts
@@ -18,6 +18,10 @@ import * as Rx from 'rx';
 
 /// <reference path="./shapes.ts" />
 import {Identity} from './shapes';
+<<<<<<< HEAD
+=======
+
+>>>>>>> fixed imports
 
 export const createdNotes = new Rx.Subject();
 export const creates = new Rx.Subject();
@@ -33,18 +37,21 @@ export const runningTotal = Rx.Observable.merge(removesCounter, createsCounter)
     .scan((acc: number, value: number) => acc + value);
 
 export const noteStack = Rx.Observable.merge(
-    createdNotes.map((x: Identity) => {
+    createdNotes.map((x: { identity: Identity, options: any }) => {
+        // opacity added here to change the property in genAnimationFunction
         return {
             create: 1,
-            name: x.name,
-            uuid: x.uuid,
+            name: x.identity.name,
+            uuid: x.identity.uuid,
+            opacity: x.options.opacity
         };
     }),
-    removes.map((x: Identity) => {
+    removes.map((x: any) => {
         return {
             create: 0,
             name: x.name,
             uuid: x.uuid,
+            opacity: x.opacity
         };
     }))
     .distinctUntilChanged()
@@ -57,6 +64,7 @@ export const noteStack = Rx.Observable.merge(
                 acc.push({
                     name: value.name,
                     uuid: value.uuid,
+                    opacity: value.opacity
                 });
             }
 
--- a/src/browser/api/notifications/observable_sequences.ts
+++ b/src/browser/api/notifications/observable_sequences.ts
@@ -18,10 +18,6 @@ import * as Rx from 'rx';
 
 /// <reference path="./shapes.ts" />
 import {Identity} from './shapes';
-<<<<<<< HEAD
-=======
-
->>>>>>> fixed imports
 
 export const createdNotes = new Rx.Subject();
 export const creates = new Rx.Subject();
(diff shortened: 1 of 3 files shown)


wenjunche:
@whyn07m3 would say:  missing tests

src/browser/api/notifications/subscriptions.ts
```diff
@@ -391,15 +391,16 @@ function genAnimationFunction(defaultTop: number, numNotes: number): (noteWin: a
     return (noteWin: any, idx: number) => {
         const {name, uuid} = noteWin;
         const identity: Identity = {name, uuid};
+        const opacity = noteWin.opacity || 1;
```
whyn07m3: :neckbeard: 
• can simply put it below into the object (optional)
• can notifications be transparent (opacity: 0)? If yes, than this won't work, just asking...

src/browser/api/notifications/observable_sequences.ts
```diff
@@ -33,18 +34,20 @@ export const runningTotal = Rx.Observable.merge(removesCounter, createsCounter)
     .scan((acc: number, value: number) => acc + value);
 
 export const noteStack = Rx.Observable.merge(
-    createdNotes.map((x: Identity) => {
+    createdNotes.map((x: NoteId) => {
         return {
             create: 1,
             name: x.name,
             uuid: x.uuid,
+            opacity: x.opacity
```
whyn07m3: Question, why does `opacity` go here and not simply into Window's options on creation in _javascript-adapter_?

tgoc99: In the notifications there is an animation that overrides the window options that was previously hard-coded to change the notification opacity to 1 over one second when the notification was first shown on the screen (see genAnimationFunction fn in subscriptions.ts file).

src/browser/api/notifications/subscriptions.ts
tgoc99: - don't really have opinion, happy to make either way my default but seems like I see it defined as variable much more often in our code
- Tried to change as little as possible, before this change all notifications were set to opacity 0 and this function would take them from 0 to 1 over time.  This way keeps that behavior constant.  Do we want people to be able to set opacity 0?

src/browser/api/notifications/observable_sequences.ts
whyn07m3: Got it, optional, but helpful would be to add a comment here explaining this

src/browser/api/notifications/subscriptions.ts
whyn07m3: I don't know how other stuff in notifications work, just if the `opacity` value here is `0` than it would be changed to `1`.

src/browser/api/notifications/shapes.ts
```diff
@@ -31,6 +31,10 @@ export interface NotificationMessage {
     data: any;
 }
 
+export interface NoteId extends Identity {
```
nicholasdgoodman: What's with the name of this interface? Seemingly the shape of this declaration is missing any kind of identity information at all...

nicholasdgoodman: Back to the naming of this interface, are we sure the interface is not defining the shape of a `Note` and not the shape of a `NoteId`?

Also, are we consistently abbreviating identity / identification as `Id` in extended implementations?

rdepena:
@tgoc99 Let's not pack the notification Identity with extra data. Identity should be limited to data that can identity the given entity.

We can pass extra options along.

rdepena:
🌳

src/browser/api/notifications/shapes.ts
rdepena: 🌳

tgoc99:
Made the fixes - @rdepena let me know if the change in `observable_sequences.ts` works for your comment.

✅ Test results:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a00a596a8ae142e4a5e37a8)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a00a5e7a8ae142e4a5e37a9)

src/browser/api/notifications/subscriptions.ts
rdepena: @tgoc99 I am with @whyn07m3, this would be better served with hasOwnProperty or something similar as `= 0 || anything` will always resolve to anything, not sure 0 is a valid value but we should code around it.

rdepena:
added a comment on @whyn07m3's observations, the other changes look good,

src/browser/api/notifications/subscriptions.ts
datamadic: 👍 on Roma's opacity. lets check if its a number here. people do weird things...

src/browser/api/notifications/observable_sequences.ts
```diff
@@ -33,18 +33,21 @@ export const runningTotal = Rx.Observable.merge(removesCounter, createsCounter)
...
             create: 1,
-            name: x.name,
-            uuid: x.uuid,
+            name: x.identity.name,
+            uuid: x.identity.uuid,
+            opacity: x.options.opacity
         };
     }),
-    removes.map((x: Identity) => {
+    removes.map((x: any) => {
```
datamadic: why are we losing the typing here?

src/browser/api/notifications/subscriptions.ts
nicholasdgoodman: do we have `isNan(...)` ? seems appropriate here:

```
const opacity = isNan(noteWin.opacity) ? 1 : noteWin.opacity;
```

datamadic:
@rdepena if the updated looks good, lets merge. Looks good to me

src/browser/api/notifications/subscriptions.ts
datamadic: I like this

datamadic: actually I dont, this transitions from 0 to 0 in the default case

tgoc99:
Made the changes to notification opacity to allow for zero opacity.  Required some changes to the [Javascript-Adapter](https://github.com/openfin/javascript-adapter/pull/352) as well.

✅ Test results:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a00da051573627ee4384249)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a00da641573627ee438424a) - (Re-ran Win10 failed tests and one passed)

PR #231, opened 2017-11-06
RUN-3505: Update spec for registry security policy

wenjunche:
Test results:
[W10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a007ad8af26a25be2ffca52)
[W7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a007ad8af26a25be2ffca52)

### DIFF
--- a/src/browser/api_protocol/api_handlers/system.js
+++ b/src/browser/api_protocol/api_handlers/system.js
@@ -17,7 +17,8 @@ let apiProtocolBase = require('./api_protocol_base.js');
 let System = require('../../api/system.js').System;
 let _ = require('underscore');
 const connectionManager = require('../../connection_manager');
-const log = require('../../log');
+import * as log from '../../log';
+import * as path from 'path';
 
 const ReadRegistryValuePolicyDelegate = {
     //checkPermissions(ApiPolicyDelegateArgs): boolean;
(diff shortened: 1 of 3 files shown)


src/browser/api_protocol/api_handlers/system.js
```diff
@@ -17,6 +17,35 @@ let apiProtocolBase = require('./api_protocol_base.js');
 let System = require('../../api/system.js').System;
 let _ = require('underscore');
 const connectionManager = require('../../connection_manager');
+const log = require('../../log');
```
whyn07m3: `import`

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -17,6 +17,35 @@ let apiProtocolBase = require('./api_protocol_base.js');
...
+    //checkPermissions(ApiPolicyDelegateArgs): boolean;
+    checkPermissions: function(args) {
+        // permissionSettings has following format
+        // { "enabled": true, "registryKeys": [ "HKEY_CURRENT_USER\\Software\\OpenFin\\RVM" ] }
+        let permitted = false; // default to false
+        if (args.payload && args.permissionSettings && args.permissionSettings.enabled === true) {
+            if (Array.isArray(args.permissionSettings.registryKeys)) {
+                let fullPath = args.payload.rootKey;
+                if (args.payload.subkey) {
+                    fullPath = fullPath.concat('\\' + args.payload.subkey);
```
whyn07m3: Can you use `path.join(...` here?

src/browser/api_protocol/api_handlers/system.js
```diff
@@ -17,6 +17,35 @@ let apiProtocolBase = require('./api_protocol_base.js');
...
+                }
+                if (args.payload.value) {
+                    fullPath = fullPath.concat('\\' + args.payload.value);
```
whyn07m3: ... and here

src/browser/api_protocol/api_handlers/system.js
nicholasdgoodman: `path.join(...)` is not suitable for joining registry key paths since registry key names can contain characters that are invalid for directory names.

For example, if a registry key name contains a forward slash, `path.join` will replace it with a backslash, and the key won't be correctly accessed.

src/browser/api_protocol/api_handlers/system.js
nicholasdgoodman: ... and here

wenjunche:
Talked to Chuck:  for any key specified in policy,  all sub-folders and values are allowed.

PR #234, opened 2017-11-07
RUN-3499 delete default max-Height/Width

BoOpenfin:
When converting window-options, delete default maxHeight and maxWidth (which is set to be -1 at base options) that the default value[ set to be no limit](https://github.com/openfin/runtime/blob/develop/atom/browser/native_window.cc#L130) at runtime correctly.

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -233,6 +233,13 @@ module.exports = {
                 newOptions.width = defaultWidth;
             }
 
+            if (maxHeight === -1) {
+                delete newOptions.maxHeight;
+            }
+            if (maxWidth === -1) {
+                delete newOptions.maxWidth;
+            }
+
             newOptions.center = newOptions.defaultCentered;
             if (!newOptions.center) {
                 newOptions.x = newOptions.defaultLeft;


wenjunche:
just remove the following from five0BaseOptions ?
        'maxHeight': -1,
        'maxWidth': -1,

BoOpenfin:
I think five0BaseOptions is like the prototype that we used to [validate the user's option](https://github.com/HadoukenIO/core/blob/0ee4f72a87e3efed0cda269e6e9b8514aa7b96d3/src/browser/convert_options.js#L177), as a result, we should have every attribute defined inside? And -1 is also [taken as the default maxHeight when refining current height](https://github.com/HadoukenIO/core/blob/0ee4f72a87e3efed0cda269e6e9b8514aa7b96d3/src/browser/convert_options.js#L230). So, delete the max-Height/Width here seems less modifications?

HarsimranSingh:
@BoOpenfin Please rebase this branch onto `develop` and kick off the automated tests for windows. Then we can merge this.

BoOpenfin:
Yes, maxHeight and maxWidth become **undefined** by default.

HarsimranSingh:
Hmm that's a valid concern @nicholasdgoodman. @BoOpenfin and I talked about this and maybe we should just handle this on the runtime side.

In the future, we should set up some sort of separation between OpenFin options and Electron options. Possible refactoring for another day...

nicholasdgoodman:
Though I personally think `undefined` is more appropriate for that which is, well, undefined; there's a slight risk a customer may have written some kind of logic which checks specifically for the value `-1`. Therefore this might be a breaking change for them.

Sad thing is there's no way to really know... save this idea for "API v2"?

HarsimranSingh:
Closing in favor of doing this on the runtime side.

PR #241, opened 2017-11-09
RUN-3536: preload script reload

wenjunche:
[JS adapter PR](https://github.com/openfin/javascript-adapter/pull/359)

Tests:

[W10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a0475041573627ee43842bd)
[W7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a0474c01573627ee43842bc)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -47,6 +47,7 @@ import { validateNavigationRules } from '../navigation_validation';
 import * as log from '../log';
 import SubscriptionManager from '../subscription_manager';
 import route from '../../common/route';
+import { getIdentifier, deletePreloadScriptState } from '../preload_scripts';
 
 const subscriptionManager = new SubscriptionManager();
 const TRAY_ICON_KEY = 'tray-icon-events';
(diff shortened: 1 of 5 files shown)


src/browser/api/application.js
```diff
@@ -837,6 +838,15 @@ Application.emitRunRequested = function(identity, userAppConfigArgs) {
...
 
+Application.reloadPreloadScripts = function(identity, preloadScripts, callback) {
+    (preloadScripts || []).map(preload => {
+        electronApp.vlog(1, `clear preload script ${getIdentifier(preload)}`);
+        deletePreloadScriptState(getIdentifier(preload));
+    });
+    System.downloadPreloadScripts(identity, preloadScripts, callback);
+
+};
+
```
nicholasdgoodman: Perhaps we should consider checking that all the requested "scripts to re-download" were previously supplied as window options. As-written, the API allows download of arbitrary files to the cache, even if they were never forward-declared. Such a feature may / may not be desirable.

Suggest checking that all requested scripts here already have an entry in the preload script cache.

nicholasdgoodman:
Another major difference between this proposed API and the existing `download*` API's is that both Asset and Runtime downloads via API will fail if the requested item is already downloaded. This is by-design.

However, this new proposed API has the intended purpose of **re-** downloading a previously downloaded file.

Unless clearly documented, this difference would represent a future "gotcha" of our API. If a `download` API allows re-download, perhaps it is gated by a setting or flag?

whyn07m3:
How are we and do we need to make sure that we only re-download what's already existed in Application and Windows...?

src/browser/api/application.js
wenjunche: if a script is not listed in window options, it will not be applied during reload

PR #257, opened 2017-11-28
RUN-3328 (phase 2) multi-runtime mesh system events

tgoc99:
Second phase implements the system events that need to be aggregated together from other runtimes.  There are timing issues that need to be fixed that result in not receiving application created or started events on a new runtime startup:   [JIRA for event timing](https://appoji.jira.com/browse/RUN-3654) 

[JS-Adapter Tests](https://github.com/HadoukenIO/js-adapter/pull/66)

✅ Test results:
• [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a1ddb039a4b9e06a9723aa6)
• [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a1ddb4c9a4b9e06a9723aa7)

### DIFF
--- a/src/browser/remote_subscriptions.ts
+++ b/src/browser/remote_subscriptions.ts
@@ -53,7 +53,13 @@ interface RemoteSubscription extends RemoteSubscriptionProps {
     unSubscriptions: Map<string, (() => void)[]>; // a map of un-subscriptions assigned to runtime versions
 }
 
-const systemEventsToIgnore: any = {
+interface SystemEventsToIgnore {
+    'idle-state-changed': boolean;
+    'monitor-info-changed': boolean;
+    'session-changed': boolean;
+}
+
+const systemEventsToIgnore: SystemEventsToIgnore = {
     'idle-state-changed': true,
     'monitor-info-changed': true,
     'session-changed': true
--- a/src/browser/remote_subscriptions.ts
+++ b/src/browser/remote_subscriptions.ts
@@ -57,6 +57,7 @@ interface SystemEventsToIgnore {
     'idle-state-changed': boolean;
     'monitor-info-changed': boolean;
     'session-changed': boolean;
+    [index: string] : boolean;
 }
 
 const systemEventsToIgnore: SystemEventsToIgnore = {
--- a/src/browser/remote_subscriptions.ts
+++ b/src/browser/remote_subscriptions.ts
@@ -53,14 +53,7 @@ interface RemoteSubscription extends RemoteSubscriptionProps {
     unSubscriptions: Map<string, (() => void)[]>; // a map of un-subscriptions assigned to runtime versions
 }
 
-interface SystemEventsToIgnore {
-    'idle-state-changed': boolean;
-    'monitor-info-changed': boolean;
-    'session-changed': boolean;
-    [index: string] : boolean;
-}
-
-const systemEventsToIgnore: SystemEventsToIgnore = {
+const systemEventsToIgnore: {[index: string]: boolean} = {
     'idle-state-changed': true,
     'monitor-info-changed': true,
     'session-changed': true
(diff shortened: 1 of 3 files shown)


src/browser/remote_subscriptions.ts
```diff
@@ -49,9 +49,16 @@ interface RemoteSubscriptionProps extends Identity {
 interface RemoteSubscription extends RemoteSubscriptionProps {
     _id: number; // ID of the subscription
     isCleaned: boolean; // helps prevents repetitive un-subscriptions and other cleanup
+    isSystemEvent?: boolean; // helps prevents repetitive un-subscriptions and other cleanup
     unSubscriptions: Map<string, (() => void)[]>; // a map of un-subscriptions assigned to runtime versions
 }
 
+const systemEventsToIgnore: any = {
+    'idle-state-changed': true,
```
whyn07m3: Should these events go to https://github.com/tgoc99/core/blob/03be9730f3b7ae18be562557808949fd55dd44a7/src/browser/api_protocol/api_handlers/mesh_middleware.ts#L20?

src/browser/remote_subscriptions.ts
```diff
@@ -212,6 +210,101 @@ function unSubscribe(subscription: RemoteSubscription, runtime: PeerRuntime) {
...
+            runtime.forEach((removeListener: any) => removeListener());
+        }
+    });
+    pendingRemoteSubscriptions.delete(subscription._id);
+}
+
+/**
+ * Subscribe to a system event in a remote runtime
+ */
+function applySystemSubscription(subscription: RemoteSubscription, runtime: PeerRuntime) {
```
whyn07m3: Can [this function](https://github.com/tgoc99/core/blob/03be9730f3b7ae18be562557808949fd55dd44a7/src/browser/remote_subscriptions.ts#L112) be reused here (a lot of code looks similar)

src/browser/remote_subscriptions.ts
```diff
@@ -49,9 +49,16 @@ interface RemoteSubscriptionProps extends Identity {
 interface RemoteSubscription extends RemoteSubscriptionProps {
     _id: number; // ID of the subscription
     isCleaned: boolean; // helps prevents repetitive un-subscriptions and other cleanup
+    isSystemEvent?: boolean; // helps prevents repetitive un-subscriptions and other cleanup
```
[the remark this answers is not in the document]
tgoc99: yep, my bad

src/browser/remote_subscriptions.ts
```diff
@@ -49,9 +49,16 @@ interface RemoteSubscriptionProps extends Identity {
...
     unSubscriptions: Map<string, (() => void)[]>; // a map of un-subscriptions assigned to runtime versions
 }
 
+const systemEventsToIgnore: any = {
```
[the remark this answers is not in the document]
tgoc99: will do

src/browser/remote_subscriptions.ts
tgoc99: In the same object?  Here we refer to system event handlers to ignore and the other file refers to system API methods.  The remote event subscriptions don't go through the same path as the other remote API calls.   They skip the middleware in that file (see ignored 'subscribe to desktop event' in that file) and get handled if necessary in https://github.com/tgoc99/core/blob/03be9730f3b7ae18be562557808949fd55dd44a7/src/browser/api_protocol/api_handlers/event_listener.js. @rdepena correct me if I'm wrong...

src/browser/remote_subscriptions.ts
tgoc99: I think its close.... I thought about trying to re-write this so that one fn could be used for both but after some advice decided against it.  Differences are: 
-  don't want cleanUpSubscription in the listener for system events
-  need to unsubscribe when runtime disconnects for system events (handled with combination of cleanUpSubscription & persistSubscription for other events)
- couple of other things but they're easy fixes

Let me know if you disagree.

whyn07m3: @tgoc99 Hard to tell for sure by taking a quick look, but if you thought of this and decided not to go that route, then okay. I'm fine with the fact that you considered it and still decided to put it in a separate function

src/browser/remote_subscriptions.ts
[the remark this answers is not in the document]
tgoc99: If I put object there, TS complains when I try to use any properties later.  I could add a whole interface but this has a very specific purpose, seems like overkill to me.  But if anyone feels strongly happy to add...

PR #272, opened 2017-12-20
RUN-3721 Fixed issue where we expected an array for preload processing.

rdepena:
Test results: 

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a3ac5156f202a5432b79f7c)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a3ac5986f202a5432b79f7f)

### DIFF
--- a/src/browser/preload_scripts.ts
+++ b/src/browser/preload_scripts.ts
@@ -81,7 +81,7 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
 export async function loadScripts(identity: Identity): Promise<PreloadScriptWithContent[]|any> {
     const options = Window.getOptions(identity);
     const { preloadScripts = [] }: {preloadScripts: Array<PreloadScript>} = convertToElectron(options);
-    const promises =  preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
+    const promises = preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
     return await Promise.all(promises);
 }
 
--- a/src/browser/preload_scripts.ts
+++ b/src/browser/preload_scripts.ts
@@ -81,7 +81,7 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
 export async function loadScripts(identity: Identity): Promise<PreloadScriptWithContent[]|any> {
     const options = Window.getOptions(identity);
     const { preloadScripts } = convertToElectron(options);
-    const promises = preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
+    const promises = preloadScripts ? preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript)) : [];
     return await Promise.all(promises);
 }
 
--- a/src/browser/preload_scripts.ts
+++ b/src/browser/preload_scripts.ts
@@ -80,8 +80,8 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
 
 export async function loadScripts(identity: Identity): Promise<PreloadScriptWithContent[]|any> {
     const options = Window.getOptions(identity);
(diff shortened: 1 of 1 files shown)


src/browser/preload_scripts.ts
```diff
@@ -81,7 +81,7 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
 export async function loadScripts(identity: Identity): Promise<PreloadScriptWithContent[]|any> {
     const options = Window.getOptions(identity);
     const { preloadScripts } = convertToElectron(options);
-    const promises = preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
+    const promises = preloadScripts ? preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript)) : [];
```
HarsimranSingh: I think the problem might lie earlier. Why doesn't this catch it https://github.com/HadoukenIO/core/blob/develop/src/browser/convert_options.js#L348

whyn07m3: IMO, we should either fix this in [here](https://github.com/HadoukenIO/core/blob/develop/src/browser/convert_options.js#L258) instead
--- or ---
do this: `const { preloadScripts = [] } = convertToElectron(options);`

rdepena: @whyn07m3 will make that change.

rdepena:
@HarsimranSingh @whyn07m3 avoided changes to convert_options as they will affect preload script inheritance, a bit more changes than I want for a hotfix.

Test results: 
[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a53cf83bd581266877d664e)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a53d05dbd581266877d664f)

src/browser/preload_scripts.ts
```diff
@@ -80,8 +80,8 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
 
 export async function loadScripts(identity: Identity): Promise<PreloadScriptWithContent[]|any> {
     const options = Window.getOptions(identity);
-    const { preloadScripts } = convertToElectron(options);
-    const promises = preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
+    const { preloadScripts = [] }: {preloadScripts: Array<PreloadScript>} = convertToElectron(options);
```
whyn07m3: Shouldn't this be `const { preloadScripts = [] }: PreloadScript[] = convertToElectron(options);`?

src/browser/preload_scripts.ts
```diff
@@ -80,8 +80,8 @@ function downloadScript(identity: Identity, preloadScript: PreloadScript): Promi
...
+    const promises =  preloadScripts.map((preloadScript: PreloadScript) => loadScript(identity, preloadScript));
```
whyn07m3: If you are going to change the above line, this line has an extra unnecessary space added and shows as a changed line

src/browser/preload_scripts.ts
rdepena: I actually prefer Array<T>

whyn07m3: But I think the type is wrong though or am I missing something?
`const { preloadScripts = [] }: { preloadScripts: Array<PreloadScript> } = convertToElectron(options);`
I think it should be:
`const { preloadScripts = [] }: Array<PreloadScript> = convertToElectron(options);`
as resulting `preloadScripts` is an array, not an object with a property...

rdepena: ahh, that is the correct syntax for destructing.

rdepena: meaning the code there is correct, preloadScripts will be an array of PreloadScript.

rdepena: https://www.typescriptlang.org/docs/handbook/variable-declarations.html#destructuring

PR #274, opened 2017-12-22
Revert back to an older version of the p2p package untill bugs are fixed.

rdepena:
Newer version of runtime-p2p has a few issues, would not affect any feature in .27 but will need fixing for .28.

Test results:
[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a3d56f36f202a5432b79f87)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a3d577f6f202a5432b79f88)


rdepena:
@HarsimranSingh should be fine to ship without it, we announced this for .28 and we already have a ticket to go back to .asar files so, not too worried here.

PR #277, opened 2017-12-28
RUN-3727 experimental application option

StevenEBarbaro:
experimental application option for shimming `--disable-init-reload` per app (process)

```
'experimental': {
  'disableInitialReload': false,
  'v2Api': false
},
```

Relates to https://github.com/openfin/runtime/pull/783

main.js already has the correct spelling
https://github.com/HadoukenIO/core/blob/develop/src/renderer/main.js#L48

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -236,11 +236,10 @@ module.exports = {
         newOptions.enableLargerThanScreen = true;
         newOptions['enable-plugins'] = true;
         newOptions.webPreferences = {
+            disableInitialReload: newOptions.experimental.disableInitialReload
             nodeIntegration: false,
             plugins: newOptions.plugins
         };
-
-        newOptions.webPreferences.disableInitialReload = newOptions.experimental.disableInitialReload;
 
         if (coreState.argo['disable-web-security'] || newOptions.webSecurity === false) {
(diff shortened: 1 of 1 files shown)


src/browser/convert_options.js
```diff
@@ -239,6 +240,12 @@ module.exports = {
             plugins: newOptions.plugins
         };
 
+        // v2Api Fallback for incorrect spelling
```
HarsimranSingh: Is there a reason to support the misspelling? This feature is going out in `8.56.27.*`, which is still in staging. Why don't we fix it there?

StevenEBarbaro: I support that 👍

src/browser/convert_options.js
```diff
@@ -239,6 +240,8 @@ module.exports = {
             plugins: newOptions.plugins
         };
 
+        newOptions.webPreferences.disableInitialReload = newOptions.experimental.disableInitialReload;
```
HarsimranSingh: Move this into https://github.com/HadoukenIO/core/pull/277/files#diff-bb76f534ea36dcd8884c45b25ab0f7b0R238

HarsimranSingh:
@StevenEBarbaro can we get more of a description of what this does and also does this need to be documented?

PR #281, opened 2018-01-04
RUN-3657 fixed bug that causes exception 

weiteho:
check if window object was destroyed before invoking the callback

✅ Test Results:
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a4d99efbd581266877d663c)
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a4d9a72bd581266877d663d)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -348,7 +348,7 @@ Application.grantWindowAccess = function() {
 };
 Application.isRunning = function(identity) {
     let uuid = identity && identity.uuid;
-    return uuid && coreState.getAppRunningState(uuid) && !coreState.getAppRestartingState(uuid);
+    return !!(uuid && coreState.getAppRunningState(uuid) && !coreState.getAppRestartingState(uuid));
 };
 Application.pingChildWindow = function() {
     console.warn('Deprecated');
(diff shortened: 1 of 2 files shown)


rdepena:
Not sure this is the right fix for this issue.

weiteho:
The fix is only to prevent the exception from happening. Very easy to reproduce if set the timeout to 10 seconds and close the window within 10 seconds.

But it might not be the best design to use a timer here, we should revisit and check why ready-to-show doesn't work as expected.

rdepena:
Hey @weiteho what I meant was that we should not silently fail, if the window has been destroyed we should still fire the callback but with additional information, alerting the requesting function that the browserWindow has been destroyed. 

So changing/adding the calling function to handle this as an error state would be preferable.

HarsimranSingh:
@weiteho can you rebase these changes onto the staging branch

weiteho:
moved checking for destroyed window in callback, rebased onto the staging branch

src/browser/api/application.js
```diff
@@ -348,7 +348,7 @@ Application.grantWindowAccess = function() {
 };
 Application.isRunning = function(identity) {
     let uuid = identity && identity.uuid;
-    return uuid && coreState.getAppRunningState(uuid) && !coreState.getAppRestartingState(uuid);
+    return !!(uuid && coreState.getAppRunningState(uuid) && !coreState.getAppRestartingState(uuid));
```
HarsimranSingh: Looks like this got carried over on the rebase. Can you remove this

weiteho: my apologies, this is reverted.

PR #282, opened 2018-01-08
RUN-3726 don't emit messages using a closed/closing WebSocket

whyn07m3:
**What happens now:**
External adapters that immediately close WebSocket connection after an API call, will cause the core to fail at attempting to send an "ack" and the logs will show an error coming from `ws` library.

**What the behavior after this change will be:**
When the core is trying to send an "ack" to an external adapter, it will only attempt to send it if the socket is open. If the socket is closing/closed it will print a debug message describing this failed attempt.

**Test results:**
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a4fea13bd581266877d663f)
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a4feafebd581266877d6640)


HarsimranSingh:
@whyn07m3 can you add a node test for this?

PR #293, opened 2018-01-18
RUN-3386 null checks to avoid exception

pbaize:
This adds the null checks to avoid the runtime exception when center clicking. It does not resolve the issue of center clicking opening a blank window.

### DIFF
--- a/src/browser/web_request_handler.ts
+++ b/src/browser/web_request_handler.ts
@@ -84,7 +84,7 @@ function beforeSendHeadersHandler(details: RequestDetails, callback: (response: 
         if (wc) {
             electronApp.vlog(1, `${moduleName}:beforeSendHeadersHandler got webcontents ${wc.id}`);
             const bw = wc.getOwnerBrowserWindow();
-            if (bw && bw.id) {
+            if (bw && typeof bw.id === 'number') {
                 const opts: Shapes.WindowOptions = coreState.getWindowOptionsById(bw.id);
                 electronApp.vlog(1, `${moduleName}:beforeSendHeadersHandler window opts ${JSON.stringify(opts)}`);
                 if (opts && opts.customRequestHeaders) {
--- a/src/browser/web_request_handler.ts
+++ b/src/browser/web_request_handler.ts
@@ -23,7 +23,7 @@ limitations under the License.
 
 const coreState = require('./core_state');
 const electronApp = require('electron').app;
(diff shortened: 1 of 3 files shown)


weiteho:
this hides the exception but not really solved the underlying issue. We should figure out what to do with center click/ctrl + click and create corresponding openfin window objects. This way we don't even need to do null check here.

pbaize:
Agreed, although we should leave the null checks in considering we sometimes set these variables to null.

src/browser/web_request_handler.ts
```diff
@@ -84,10 +84,10 @@ function beforeSendHeadersHandler(details: RequestDetails, callback: (response:
         if (wc) {
             electronApp.vlog(1, `${moduleName}:beforeSendHeadersHandler got webcontents ${wc.id}`);
             const bw = wc.getOwnerBrowserWindow();
-            if (bw) {
+            if (bw && bw.id) {
```
HarsimranSingh: Is there ever a chance that the BrowserWindow id is going to be `0`? Maybe safer to check `typeof`

wenjunche:
Manual tests on disableModifiedClicks ?

src/renderer/api-decorator.js
```diff
@@ -333,6 +333,25 @@ limitations under the License.
...
+            const mightOpenNewWindow = tag === 'A' || tag === 'IMG';
+            if (mightOpenNewWindow && modifiedClick) {
+                e.preventDefault();
+            } else if (modifiedClick && (tag === 'BUTTON' || tag === 'INPUT')) {
+                if (e.target.type === 'submit') {
+                    e.preventDefault();
+                }
+
+            }
+        });
```
whyn07m3: Code looks good as it is, just a refactored suggestion below
```
const { tagName, type } = e.target;
const modifiedClick = e.shiftKey || e.metaKey || e.ctrlKey || e.altKey;
const shouldPreventDefault =
    tagName === 'A' || tagName === 'IMG'
    ||
    type === 'submit' && (tagName === 'BUTTON' || tagName === 'INPUT');

if (modifiedClick && shouldPreventDefault) {
    e.preventDefault();
}
```

pbaize:
Test results: https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a6a3dda3f32861e82b77c84

PR #303, opened 2018-01-29
RUN-3783 Window Group Resize 

tgoc99:
Per [this JIRA](https://appoji.jira.com/browse/RUN-3783) and convo with @datamadic, made re-sizing of grouped windows work differently.  The windows are now resizing any other windows that share the same border.  I.e. if the right border of a window is dragged to the right, that window will get larger and any other windows that share that border (+/- 4 pixels) will also get larger/smaller depending on whether that border represents their left (gets smaller) or right (gets larger) border.   I think this works in more instances than before (would previously just move (not resize) any windows to the right of the right border by the amount of the increase in width) and definitely makes more sense for snap & dock.  

With this behavior - changing any internal borders keeps the overall group the same size.  Changing the outer border can change the area of the overall group.  

Amended [this test](https://testing-dashboard.openfin.co/#/app/tests/5a68f1f63f32861e82b77c6e/edit) to reflect new behavior. 

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a6f9d7f3f32861e82b77cbf)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a6f9e473f32861e82b77cc0)


src/browser/bounds_changed_state_tracker.js
```diff
@@ -21,6 +21,7 @@ let windowTransaction = require('electron').windowTransaction;
 
 let _ = require('underscore');
 let animations = require('./animations.js');
+let clipBounds = require('./clip_bounds.js').default;
```
whyn07m3: This should `import clipBounds from './clip_bounds';`

src/browser/bounds_changed_state_tracker.js
```diff
@@ -215,36 +276,25 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
...
+                        flags = noZorder + noSize + noActivate;
+                    }
 
                     WindowGroups.getGroup(groupUuid).filter((win) => {
                         win.browserWindow.bringToFront();
                         return win.name !== name;
                     }).forEach((win) => {
-                        let { x, y, width, height } = win.browserWindow.getBounds();
+                        let winBounds = win.browserWindow.getBounds();
+                        let { x, y, width, height } = (changeType === 1) ? handleGroupedResize(boundsCompare, delta, cachedBounds, win): winBounds;
```
whyn07m3: Both should be `const`s?

src/browser/bounds_changed_state_tracker.js
```diff
@@ -264,12 +314,13 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
                             if (win.browserWindow.isMaximized()) {
                                 win.browserWindow.unmaximize();
                             }
-                            wt.setWindowPos(hwnd, { x, y, flags });
+                            let [w, h] = [width, height];
```
whyn07m3: `const`

src/browser/bounds_changed_state_tracker.js
tgoc99: winBounds yes other no - may be redefined below.  Requested Changes made.

tgoc99:
Requested changes made - lets turned into consts

PR #306, opened 2018-01-31
RUN-3371 noisy and fatal invalid preload scripts window options

whyn07m3:
**[Companion PR](https://github.com/openfin/javascript-adapter/pull/387)**

**About**
This PR makes the improper use of preload scripts a "noisy" thing.

*Examples of what's changing:*

1. Using deprecated `preload` option in app's **manifest** shows this in the `debug.log`:
`[preloadScripts] [uuid]-[name]: 'preload' option is deprecated, use 'preloadScripts' instead`

2. Using a falsy value for `preloadScripts` instead of an empty array in app's **manifest** shows this in the `debug.log`:
`[preloadScripts] [uuid]-[name]: Consider using an empty array with 'preloadScripts', instead of a falsy value`

3. Using invalid value (e.g. `true`) for either `preload` or `preloadScripts` in app's **manifest** will throw a fatal error.
image

**Manual test**
[Fatal error for invalid preload scripts shape](https://github.com/openfin/test_apps/pull/21)

**Automatic tests**
*Listed in the companion javascript-adapter PR...*

**Test results**
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a71f1df3f32861e82b77ced)
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a71f2a73f32861e82b77cee)


index.js
```diff
@@ -681,3 +683,42 @@ function registerShortcuts() {
 function needsCrashReporter(argo) {
     return !!(argo['diagnostics'] || argo['enable-crash-reporting']);
 }
+
+function validatePreloadScripts(options) {
```
rdepena: Any reason this is not in preload_scripts.ts and imported into main?

whyn07m3: @rdepena I initially put it in `preload_scripts.ts` (as well as the normalization function that is located in `convert_options` now), but importing this/these creates circular dependencies, so had to move it out of `preload_scripts.ts`

rdepena: gotcha

PR #314, opened 2018-02-08
RUN-3816: Make runtimes v2Api aware

StevenEBarbaro:


src/browser/convert_options.js
```diff
@@ -241,7 +241,8 @@ module.exports = {
             nodeIntegration: false,
             plugins: newOptions.plugins,
             preload: (!useNodeInRenderer ? noNodePreload : ''),
-            sandbox: !useNodeInRenderer
+            sandbox: !useNodeInRenderer,
+            v2Api: newOptions.experimental.v2Api
```
HarsimranSingh: It doesn't seem like anything on the runtime side is expecting this? Is there planned work?

PR #341, opened 2018-02-28
RUN-3839 Removed the usage of the browser side API to generate GUIDs, replaced it with the render side getGUID API

rdepena:
Consumed by the adapters new and old. 

### Tests:
[Win 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a9713016a994a57faa5c99a)
[Win 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5a97154f6a994a57faa5c99b)

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -109,10 +109,6 @@ limitations under the License.
         return socketServerState;
     }
 
-    function generateGuidSync() {
-        return syncApiCall('generate-guid');
-    }
-
     function convertOptionsToElectronSync(options) {
         return syncApiCall('convert-options', options);
     }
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -416,7 +412,7 @@ limitations under the License.
                     onContentReady(window, cb);
                 }
             },
-            getUuid: generateGuidSync,
+            getUuid: process.getGUID,
```
HarsimranSingh: Where is this coming from? A quick search in core didn't bring anything up.

rdepena: @StevenEBarbaro Added this to the runtime recently: 

https://github.com/openfin/runtime/blob/develop/atom/renderer/atom_sandboxed_renderer_client.cc#L93

HarsimranSingh: 👍

PR #357, opened 2018-03-19
RUN-3895 remove optional dependencies from npm-shrinkwrap to fix build

HarsimranSingh:
Removed runtime-p2p from npm-shrinkwrap.json since optionalDependencies don't seem to play well with shrinkwrapping.

Also changed openfin-sign to use the GitHub URL syntax over ssh.


rdepena:
👍

PR #373, opened 2018-04-10
RUN-4008: Send parent config url to RVM on app started event

patpao:
[JIRA](https://appoji.jira.com/browse/RUN-4008)

We want to send the parent app's config url to the RVM on the app-started event, since it will be helpful for us to tie together different records when looking at licensing data.

Tests:
[W7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5acbcc434ecc2a37d5a48439)
[W10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5acbcdac4ecc2a37d5a4843a)

I investigated the test failures and they appear to be flaky tests, unrelated to my changes. I reran each failure individually and observed the following:

* Window.navigateBack passes when I reran it with my asar
* Window.getSnapshot, Window.getSnapshot T2, and Notification.close all fail on regular canary, without my asar

Not sure if those are known issues or not

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -500,6 +500,13 @@ function run(identity, mainWindowOpts, userAppConfigArgs) {
 
         coreState.setLicenseKey({ uuid }, licenseKey);
 
+        const parentUuid = appState.parentUuid || null;
+
+        let parentConfigUrl = null;
+        if (parentUuid) {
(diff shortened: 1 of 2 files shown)


wenjunche:
probably should call core_state.getUuidBySourceUrl to get manifest URL for the 'root' app

whyn07m3:
@wenjunche look like the function `getConfigUrlByUuid` will get config URL of the root app, or is it different?

PR #374, opened 2018-04-11
RUN-3988 Multi-Runtime Join Group_1

tgoc99:
Currently the runtime where the request is made is the one handling the join group.  The middleware checks to see if either of the windows in another runtime, if so - it gets the nativeId from the other runtime, creates an external window (identity that makes the join group api call is the parent window), and subscribes to the necessary events, then uses the created external window to handle group movements.  

@wenjunche or whoever else knows external windows - please see line 500 in window.js.  I am skipping the setTaskbar fn for the external windows created here since it was erroring out (errors on l.1823 - `browserWindow.setIcon(icon);`, brings down runtime).  Not sure why it was freezing up the app, icon argument looked OK.

There is still a lot to do but this PR works to joingroup when called from an OF app:
- Currently cannot handle joinGroup calls from external applications (can't test w/ js-adapter...)
- close window / ungroup unhandled
- needs more testing with various runtime options (3 runtimes that matter each call: called from, window 1, window 2)

Tests: 
[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ace55104ecc2a37d5a48452)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ace55804ecc2a37d5a48453)

### DIFF
--- a/src/browser/api_protocol/api_handlers/mesh_middleware.ts
+++ b/src/browser/api_protocol/api_handlers/mesh_middleware.ts
@@ -157,7 +157,8 @@ const handleExternalWindow = async (identity: Identity, toGroup: Identity) => {
         const childWindowOptions = {
             uuid: identity.uuid,
             name,
-            hwnd: hwnd.data
+            hwnd: hwnd.data,
+            meshJoinGroup: true
         };
         const parent = coreState.getWindowByUuidName(identity.uuid, identity.name);
         const parentId = parent && parent.browserWindow && parent.browserWindow.id;
--- a/src/browser/api_protocol/api_handlers/mesh_middleware.ts
+++ b/src/browser/api_protocol/api_handlers/mesh_middleware.ts
@@ -18,7 +18,7 @@ import { MessagePackage } from '../transport_strategy/api_transport_base';
 import * as log from '../../log';
 import { default as connectionManager } from '../../connection_manager';
 import ofEvents from '../../of_events';
-import { Identity } from 'hadouken-js-adapter/out/src/identity';
+import { Identity } from '../../../shapes';
 import { BrowserWindow } from 'electron';
 import { Window } from '../../api/window';
 import route from '../../../common/route';
@@ -136,9 +136,12 @@ function sendMessageMiddleware(msg: MessagePackage, next: () => void) {
 
 const handleExternalWindow = async (identity: Identity, toGroup: Identity) => {
     const { uuid, name } = toGroup;
+    if (coreState.getWindowByUuidName(identity.uuid, name)) {
+        return 'registered';
+    }
     try {
         const getHwndMessage = {
-            action: 'get-hwnd',
+            action: 'get-window-native-id',
             payload: {
                 uuid,
                 name
@@ -162,7 +165,7 @@ const handleExternalWindow = async (identity: Identity, toGroup: Identity) => {
         }
         Window.create(childId, childWindowOptions);
 
-        const teardownlistener = async () => {
+        const teardownListener = async () => {
             const subscription: RemoteSubscriptionProps = {
                 uuid,
                 name,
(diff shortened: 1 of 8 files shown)


src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -128,6 +134,84 @@ function sendMessageMiddleware(msg: MessagePackage, next: () => void) {
     }
 }
 
+const handleExternalWindow = async (identity: Identity, toGroup: Identity) => {
```
whyn07m3: Should type out return, like `: Promise<HWND|Error>`...

src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -128,6 +134,84 @@ function sendMessageMiddleware(msg: MessagePackage, next: () => void) {
...
+        Window.create(childId, childWindowOptions);
+
+        return hwnd;
+    } catch (e) {
+        log.writeToLog('info', e);
+        return e;
+    }
+};
+
+async function meshJoinWindowGroupMiddleware(msg: MessagePackage, next: (locals?: object) => void) {
```
whyn07m3: Should type out return, like `: Promise<void>`...

src/browser/api_protocol/api_handlers/middleware_entity_existence.ts
```diff
@@ -23,6 +23,7 @@ import { windowApiMap } from './window.js';
 const apisToIgnore = new Set([
     // Application
     'create-application',
+    'join-window-group',
```
whyn07m3: :neckbeard: :neckbeard: :neckbeard: ⚠️ This **brakes** alphabetical order 😱

src/browser/api_protocol/api_handlers/mesh_middleware.ts
tgoc99: :+1:

src/browser/api_protocol/api_handlers/middleware_entity_existence.ts
tgoc99: I brake for breaking the order!  🛑

src/browser/api_protocol/api_handlers/mesh_middleware.ts
tgoc99: :+1:

src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -128,6 +134,85 @@ function sendMessageMiddleware(msg: MessagePackage, next: () => void) {
     }
 }
 
+const handleExternalWindow = async (identity: Identity, toGroup: Identity): Promise<string|void> => {
```
whyn07m3: Shouldn't this be `: Promise<string|Error>`?

src/browser/api_protocol/api_handlers/mesh_middleware.ts
```diff
@@ -128,6 +134,85 @@ function sendMessageMiddleware(msg: MessagePackage, next: () => void) {
...
+                name
+            }
+        };
+        const id = await connectionManager.resolveIdentity({uuid});
+        const hwnd = await id.runtime.fin.System.executeOnRemote(identity, getHwndMessage);
+
+        const childWindowOptions = {
+            uuid: identity.uuid,
+            name,
+            hwnd: hwnd.data,
```
HarsimranSingh: So to clarify some terminology, an "external window" is a window that wasn't created by OpenFin. This could something like a Google Chrome window or an Excel window.

Now if in a multi-runtime setup you're trying to `joinGroup` with another OpenFin window, you shouldn't be creating a new BrowserWindow. What happens as a result is that now there are two BrowserWindows per native window, which can cause some weird stuff to happen, as you pointed out.

What you should be doing to refer to another OpenFin window is using `wrap`. And from there you can use `getNativeId` to get the `hwnd` to use in Window Transactions for group moving.

src/browser/api_protocol/api_handlers/mesh_middleware.ts
tgoc99: moved error handling out of this fn, think it is much cleaner

PR #380, opened 2018-04-18
Feature/RUN-3958 window.open with options

rdepena:
### Features supported

We support a subset of [features as described by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/open), features not supported are either not relevant to OpenFin windows or non standard. The list of features we support are:

* [height](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#height)
* [width](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#width)
* [top](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#top)
* [left](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#left)
* [centerscreen](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#centerscreen)
* [resizable](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#resizable)
* [chrome](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#chrome)
* [alwaysRaised](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#alwaysRaised)
* [minimizable](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#minimizable)

### Window Names

We will try to use the Window name given to `window.open`, but if this uuid/name combo already exists we will generate a random name. This is suboptimal as the developer will get a different result without any prompt but we cannot produce an error as one of the use cases for this changeset is running code that is not OpenFin aware in the runtime and we cannot expect handling of failure at this stage.

### Flags

This behaviour will be on by default but can be turned off by a flag `--native-window-open`, similar to the [electron nativeWindowOpen option](https://github.com/electron/electron/blob/e73326a324b3e6998bad1517b98193be989b28ee/lib/renderer/window-setup.js#L117)

### Tests

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ad73eee4ecc2a37d5a4848f)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ad73f4f4ecc2a37d5a48490)

I am writing new tests now.

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -463,8 +463,7 @@ limitations under the License.
         features = features.trim();
         let featuresObj = {};
         features.split(',').forEach(item => {
-            const splitItem = item.split('=');
-            const [name, value] = splitItem;
+            const [name, value] = item.split('=');
 
             switch (name) {
                 case 'height':
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -52,6 +52,7 @@ limitations under the License.
     // entity information accordingly
     const frameInfo = frames.find(e => e.frameRoutingId === renderFrameId);
     const entityInfo = frameInfo || glbl.__startOptions.entityInfo;
+    const decorateOpen = !runtimeArguments.includes('--native-window-open');
 
     let getOpenerSuccessCallbackCalled = () => {
         customData.openerSuccessCalled = customData.openerSuccessCalled || false;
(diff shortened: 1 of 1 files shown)


rdepena:
@datamadic window.open will continually open windows in OpenFin, this does not change that behavior.

top=foo will eventually become top=0 after we create the Window.

flag should not be experimental as it those are not inherited and this would serve as a kill switch.

src/renderer/api-decorator.js
```diff
@@ -415,6 +445,61 @@ limitations under the License.
...
+            default:
+                return true;
+        }
+    }
+
+    //We need to do this as all values are text and convertToElectron does not handle type changes only name translation.
+    function featuresToOptionsObj(features) {
+        features = features.split(' ').join('');
+        let featuresObj = {};
+        features.split(',').forEach(item => {
```
pbaize: :neckbeard: function would be much shorter if you reduced the array into featuresObj with returns in the switch statement.

PR #389, opened 2018-04-23
RUN-4041 use an array over new JS map

HarsimranSingh:
Something wasn't playing nicely on http://get.adobe.com/flashplayer/about/ with `new Map()`. Replacing it with an array of arrays seemed to do the trick.

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -223,22 +223,22 @@ limitations under the License.
         const { uuid, name, parent, entityType } = entityInfo;
         const winIdentity = { uuid, name };
         const parentFrameName = parent.name || name;
-        const eventMap = new Map();
+        const eventMap = [];
 
-        eventMap.set(`window/initialized/${uuid}-${name}`, winIdentity);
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -223,22 +223,22 @@ limitations under the License.
...
+        eventMap.push([`window/connected/${uuid}-${name}`, winIdentity]);
+        eventMap.push([`window/frame-connected/${uuid}-${parentFrameName}`, {
             frameName: name,
             entityType
-        });
-        eventMap.set(`frame/connected/${uuid}-${name}`, winIdentity);
+        }]);
+        eventMap.push([`frame/connected/${uuid}-${name}`, winIdentity]);
 
         asyncApiCall('raise-many-events', [...eventMap]);
```
datamadic: https://github.com/HadoukenIO/core/pull/389/files#diff-afd6093bc77cc147dad3d61fcfdec439R243 is diff if this is an array no? have we check where this is consumed? 

````
var a = [{s:1, d:2}]
[...a]
// not the same as 
var m = new Map([['s', 1], ['d',2]])
[...m]
````

PR #399, opened 2018-05-01
Feature/RUN-4049 node true sad

rdepena:
The process.eval behaves in unexpected ways, reverting back to traditional eval until we can figure out what the issue is.

@StevenEBarbaro


StevenEBarbaro:
🤠

PR #402, opened 2018-05-03
RUN-3988 mesh join group 1st cut

tgoc99:
*** DO NOT REVIEW ***

Currently the runtime where the request is made is the one handling the join group unless the target 
window is already in a group in another runtime. The middleware checks to see if either of the windows in another runtime, if so - it gets the nativeId from the other runtime, creates an external window (identity that makes the join group api call is the parent window), and subscribes to the necessary events, then uses the created external window to handle group movements.  

It is currently creating the externalWindow as window underneath the application that called joinGroup.  This may need to change.  

Known Issues / Further Work:
- getGroup needs to be fixed to work cross runtime and to return the correct window (not external window)
- Some issues delegating the parent app when using 3+ runtimes (need to solve getGroup to fix)
- Grouping breaks when parent application is closed (app where joinGroup is called from)
- mergeGroups has not been addressed
- does not work when called from an external adapter (or anything without a browserWindow)
- Not all use cases with 3+ runtimes have been tested
- API calls have not been fully tested in all cases (window hidden, minimized, etc.)
- If called with two windows with the same name from two different apps (different UUIDs) the second will fail
- close-requested has not been tested

It currently cannot be tested with the js-adapter since calls from an external-adapter fail.  This is all behind a flag (`mesh-join-group`).

Tests: 
[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ace55104ecc2a37d5a48452)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ace55804ecc2a37d5a48453)

### DIFF
--- a/src/browser/api_protocol/api_handlers/mesh_window_grouping.ts
+++ b/src/browser/api_protocol/api_handlers/mesh_window_grouping.ts
@@ -1,0 +1,239 @@
+/*
+Copyright 2018 OpenFin Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import { default as RequestHandler } from '../transport_strategy/base_handler';
+import { MessagePackage } from '../transport_strategy/api_transport_base';
+import * as log from '../../log';
+import { default as connectionManager } from '../../connection_manager';
+import ofEvents from '../../of_events';
+import { Identity, OpenFinWindow, APIPayloadAck, EventPayload } from '../../../shapes';
+import { BrowserWindow } from 'electron';
+import route from '../../../common/route';
+import { addRemoteSubscription, RemoteSubscriptionProps } from '../../remote_subscriptions';
+
+const { Window } = require('../../api/window');
+const coreState = require('../../core_state');
+const WindowGroup = require('../../window_groups.js');
+
+
+const isLocalUuid = (uuid: string): Boolean => {
+    const externalConn = coreState.getExternalAppObjByUuid(uuid);
+    const app = coreState.getAppObjByUuid(uuid);
+
+    return externalConn || app ? true : false;
+};
+
+const handleExternallyGroupedLocalWindow = async (ofWindow: OpenFinWindow, msg: MessagePackage) => {
+    const externalParentUuid = ofWindow && ofWindow.meshGroupUuid;
+    const { data, ack, nack } = msg;
+    const action = data && data.action;
+
+    const id = await connectionManager.resolveIdentity({uuid: externalParentUuid});
+    const message = {...data };
+    if (action !== 'leave-window-group') {
+        message.payload = {...message.payload, groupingUuid: externalParentUuid, groupingWindowName: externalParentUuid };
+    }
+    id.runtime.fin.System.executeOnRemote({ uuid: externalParentUuid, name: externalParentUuid }, message)
+    .then(ack)
+    .catch(nack);
+    return;
+};
+
+const meshJoinGroupEvents = (identity: Identity, grouping: Identity): void => {
+    const unsubscriptions: (Promise<() => void>)[] = [];
+    const localUnsubscriptions: (() => void)[] = [];
+    const eventMap: any = {
+        'begin-user-bounds-changed': 'begin-user-bounds-change',
+        'end-user-bounds-changed': 'end-user-bounds-change',
+        'bounds-changing': 'moving',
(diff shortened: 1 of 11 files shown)


src/browser/api_protocol/api_handlers/mesh_window_grouping.ts
```diff
@@ -0,0 +1,238 @@
...
+    const { uuid, name } = toGroup;
+    const parentUuid = parent && parent.uuid;
+
+    const meshJoinGroupMessage = {
+        action: 'set-mesh-group-uuid',
+        payload: {
+            uuid,
+            name,
+            meshGroupUuid: parentUuid
+        }
```
tgoc99: probably not necessary -  just need to get hwnd

src/browser/api_protocol/api_handlers/mesh_window_grouping.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    // below call sets meshGroupUuid on the runtime of the target and returns the nativeId of the window
+    const nativeIdResponse = await setMeshJoinGroup(identity, toGroup);
+
+    const childWindowOptions = {
+        uuid: identity.uuid,
+        name,
+        hwnd: nativeIdResponse.data,
+        meshJoinGroupIdentity: { uuid, name }
+    };
```
tgoc99: if you set the hwnd field, native knows to create external window

src/browser/api_protocol/api_handlers/mesh_window_grouping.ts
```diff
@@ -0,0 +1,238 @@
...
+
+const isLocalUuid = (uuid: string): Boolean => {
+    const externalConn = coreState.getExternalAppObjByUuid(uuid);
+    const app = coreState.getAppObjByUuid(uuid);
+
+    return externalConn || app ? true : false;
+};
+
+const handleExternallyGroupedLocalWindow = async (ofWindow: OpenFinWindow, msg: MessagePackage) => {
+    const externalParentUuid = ofWindow && ofWindow.meshGroupUuid;
```
tgoc99: whole fn likely not necessary with new arch

src/browser/api_protocol/api_handlers/middleware_entity_existence.ts
```diff
@@ -32,6 +33,11 @@ const apisToIgnore = new Set([
     'window-is-notification-type'
 ]);
 
+if (coreState.argo && coreState.argo['mesh-join-group']) {
+    apisToIgnore.add('join-window-group');
```
tgoc99: make sure it doesnt get sent to other RT to be handled

src/browser/api_protocol/api_handlers/window.js
```diff
@@ -387,6 +389,16 @@ function getWindowNativeId(identity, message, ack) {
...
 
+function setMeshWindowGroupUuid(identity, message, ack) {
+    const { payload } = message;
+    const uuid = payload && payload.meshGroupUuid;
+    const dataAck = _.clone(successAck);
+    const windowIdentity = apiProtocolBase.getTargetWindowIdentity(payload);
+
+    dataAck.data = Window.setWindowGroupUuid(windowIdentity, uuid);
+    ack(dataAck);
+}
```
tgoc99: prob not necessary - was in order to delegate moves to certian runtime

src/browser/bounds_changed_state_tracker.js
```diff
@@ -291,7 +290,7 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
                 }
 
                 groupLeader = WindowGroupTransactionTracker.getGroupLeader(groupUuid) || {};
-                if (groupLeader.name === name && checkTrackingApi(groupLeader)) {
+                if (groupLeader.name === name && checkTrackingApi(groupLeader) && !ofWindow.skipNext) {
                     var delta = getBoundsDelta(currentBounds, cachedBounds);
```
tgoc99: hacky way to get around the infinite loop that occurred with programmatic moves....

src/browser/external_window_event_adapter.js
```diff
@@ -37,120 +38,127 @@ class ExternalWindowEventAdapter {
...
-                if (currState === 'minimized') {
-                    browserWindow.emit('minimize');
-                } else if (currState === 'maximized') {
-                    browserWindow.emit('maximize');
-                } else {
-                    browserWindow.emit('restore');
+        const eventMap = {
+            'focus': () => {
+                if (!browserWindow.isDestroyed()) {
+                    browserWindow.emit('focus');
```
tgoc99: I think this has been fixed - external events were not being cleaned up.  The only changes to this file are unsubscribe

src/browser/remote_subscriptions.ts
```diff
@@ -149,6 +146,9 @@ async function applyRemoteSubscription(subscription: RemoteSubscription, runtime
     unSubscriptions.get(runtimeKey).push(() => {
         classEventEmitter.removeListener(eventName, listener);
     });
+
+    // Subscribe to an event on a remote runtime
+    classEventEmitter[listenType](eventName, listener);
 }
```
tgoc99: not sure why i thought this fixed that famous issue....

PR #407, opened 2018-05-14
RUN-4059 - Additional payload added to System.getAllWindows

tgoc99:
Edit: removed incorrect language around frame options 

Per customer request, we are adding state ('normal', 'minimized', 'maximized') and isShowing (boolean) to the System.getAllWindows call.  

[Testrunner Test:](https://testing-dashboard.openfin.co/#/app/tests/5af9dd2b4ecc2a37d5a48649/edit) that tests new payload

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5af9de7f4ecc2a37d5a4864c)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5af9dfb74ecc2a37d5a4864d)

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1081,7 +1081,6 @@ Window.disableFrame = function(identity) {
     }
 
     browserWindow.setUserMovementEnabled(false);
-    setOptOnBrowserWin('frame', false, browserWindow);
 };
 
 Window.embed = function(identity, parentHwnd) {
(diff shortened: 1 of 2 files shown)


src/browser/api/window.js
```diff
@@ -1081,6 +1081,7 @@ Window.disableFrame = function(identity) {
     }
 
     browserWindow.setUserMovementEnabled(false);
+    setOptOnBrowserWin('frame', false, browserWindow);
```
HarsimranSingh: It's a little confusing, but the window option `frame` and `disableFrame/enableFrame` aren't referring to the same thing. `disableFrame/enableFrame` should be renamed to something like `setMovable/setUnmovable`. These api calls toggle whether the user can move the window via mouse.

tgoc99:
Removed incorrect frame changes per @HarsimranSingh

PR #410, opened 2018-05-14
RUN-4104: Add ability to mask RVM payloads; do so for application-log

anguslim:
[JIRA](https://appoji.jira.com/browse/RUN-4104)

Also, if the payload is to be masked, it's not written to the log.

[Runtime PR](https://github.com/openfin/runtime/pull/996)

[Test results](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5afb140f4ecc2a37d5a48677)

### DIFF
--- a/src/browser/rvm/rvm_message_bus.ts
+++ b/src/browser/rvm/rvm_message_bus.ts
@@ -319,10 +319,6 @@ export class RVMMessageBus extends EventEmitter  {
 
         this.recordCallbackInfo(callback, timeToLive, envelope);
 
-        if (!maskPayload) {
-            log.writeToLog(1, envelope, true);
-        }
-
         return this.transport.publish(envelope, maskPayload);
     }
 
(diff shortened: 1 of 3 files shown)


anguslim:
No particular reason; I just thought it was too specific for the generic `RvmMsgBase` object. Do we expect to suppress logs for messages in the future (in which case it'd make more sense in `RvmMsgBase`)?

datamadic:
Is the RVM using it for something, or is it just used in the core?

anguslim:
It's only used in the core and runtime. The RVM has to receive the actual payload since it's the actual entity writing to disk.

anguslim:
That's true, the field is unnecessarily added to the payload. It's now a separate argument.

datamadic:
@anguslim can you post updated test results? 

@whyn07m3 looks good?

whyn07m3:
@datamadic oh, sorry, don't have rights to review

PR #420, opened 2018-05-18
RUN-4143 - Emit Begin User Bounds Changing

tgoc99:
Emit the `begin-user-bounds-changing` event for use in layouts / snap-and-dock.  

See [javascript-adapter](TBD) PR for documentation.

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5aff1c294ecc2a37d5a486df)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5aff1db84ecc2a37d5a486e0)

### DIFF
--- a/src/browser/bounds_changed_state_tracker.js
+++ b/src/browser/bounds_changed_state_tracker.js
@@ -446,7 +446,8 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
             const payload = { uuid, name, top: cachedBounds.y, left: cachedBounds.x };
             ofEvents.emit(route.window('begin-user-bounds-changing', uuid, name), Object.assign(payload, cachedBounds));
         },
-        'end-user-bounds-change': () => {
+        'end-user-bounds-change': (p) => {
+            ofEvents.emit(route.window('end-user-bounds-changing', uuid, name), { hi: 'hi' });
             setUserBoundsChangeActive(false);
             handleBoundsChange(false, true);
         },
--- a/src/browser/bounds_changed_state_tracker.js
+++ b/src/browser/bounds_changed_state_tracker.js
@@ -447,8 +447,10 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
             ofEvents.emit(route.window('begin-user-bounds-changing', uuid, name), Object.assign(payload, cachedBounds));
         },
         'end-user-bounds-change': (p) => {
-            ofEvents.emit(route.window('end-user-bounds-changing', uuid, name), { hi: 'hi' });
             setUserBoundsChangeActive(false);
+            const bounds = getCurrentBounds();
+            const payload = { uuid, name, top: bounds.y, left: bounds.x };
+            ofEvents.emit(route.window('end-user-bounds-changing', uuid, name), Object.assign(payload, bounds));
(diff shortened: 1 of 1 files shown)


src/browser/bounds_changed_state_tracker.js
```diff
@@ -440,6 +442,9 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
     var _listeners = {
         'begin-user-bounds-change': () => {
             setUserBoundsChangeActive(true);
+            const cachedBounds = getCachedBounds();
```
datamadic: Do they need / did they ask for the cached bounds somewhere?

pbaize: It was considered a "nice to have".

tgoc99: Yes, I believe the idea is to have the starting point before any user-generated moves.  This would probably be a good place to put the modifier keys (shift) as well but that will require native work.

datamadic: will this fire for every window in the group?

tgoc99:
@HarsimranSingh - I thought about the end event, but the request was only for the beginning event and bounds (which you couldn't get with our current API - the first `bounds-changing` event doesn't have the pre-move bounds) and after any `begin-user-bounds-changing` event, the next `bounds-changed` event will signify the end and gives all the necessary info.

Don't feel strongly but don't think it makes sense to add `end-user-bounds-changing` if we are not documenting publicly.  Happy to be overruled if you/anyone feels differently with that background or there is a use case I am missing.  

Will close the javascript-adapter PR with the documentation.

datamadic:
FWIW, I think that if we are adding an event and telling clients that they can use it we should document it. I don't feel strongly about the symmetry on the close event, although if we dont have an `end-user-...` event than in the documentation for this one we should mention that the correct event to use if you want the end event is the existing `boudns-changed`

tgoc99:
Makes sense @datamadic, I hadn't given @HarsimranSingh the context that this was a customer request.  I will re-open the adapter PR and add the language re: ending with bounds-changed.

pbaize:
@tgoc99 I think adding the `end` event makes a lot of sense, not only for symmetry but from a layouts perspective (being able to toggle current moving window). I could also see us using it to dynamically toggle listeners for bounds changing for performance reasons.

src/browser/bounds_changed_state_tracker.js
```diff
@@ -440,9 +442,15 @@ function BoundsChangedStateTracker(uuid, name, browserWindow) {
...
+            const payload = { uuid, name, top: cachedBounds.y, left: cachedBounds.x };
+            ofEvents.emit(route.window('begin-user-bounds-changing', uuid, name), Object.assign(payload, cachedBounds));
         },
-        'end-user-bounds-change': () => {
+        'end-user-bounds-change': (p) => {
```
HarsimranSingh: where is the `p` getting used?

tgoc99: my mistake, removed

tgoc99:
New [manual test PR](https://github.com/openfin/test_apps/pull/23)

PR #421, opened 2018-05-18
Run 4129/services port discovery

pbaize:


pbaize:
Testing is done through the js-adapter, pr for that incoming, just waiting on rvm changes.

PR #423, opened 2018-05-21
RUN-4084 preloadScripts & icons cookies fix

whyn07m3:
**About**
These changes make resource fetcher use session cookies when its making requests

**Tests** 
🚫 Can't add automated tests 👎 
🚫 Adding a manual test will take too much time 👎 

**Automated test results**
✅ [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5aff22d94ecc2a37d5a486e5)
✅ [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5aff24374ecc2a37d5a486e6)


src/browser/cached_resource_fetcher.ts
```diff
@@ -226,14 +229,22 @@ function download(fileUrl: string, filePath: string): Promise<any> {
...
         });
 
         request.once('error', (err: Error) => {
             reject(err);
         });
 
-        request.end();
+        if (session) {
+            session.cookies.get({}, (error, cookies) => {
+                const cookiesNameValue = cookies.map((e) => `${e.name}=${e.value}`);
```
datamadic: What if there is an error? Should we be handling, rejecting, or we don't care?

src/browser/api/application.js
```diff
@@ -707,6 +693,17 @@ function run(identity, mainWindowOpts, userAppConfigArgs) {
     coreState.setAppRunningState(uuid, true);
 
     ofEvents.emit(route.application('started', uuid), { topic: 'application', type: 'started', uuid });
+
+    const { preloadScripts } = mainWindowOpts;
+    const loadUrl = () => app.mainWindow.loadURL(app._options.url);
```
datamadic: Not sure it matters, but this moves the app started event before the loading of the url, was that intentional?

src/browser/api/application.js
```diff
@@ -707,6 +693,17 @@ function run(identity, mainWindowOpts, userAppConfigArgs) {
...
+
+    if (appWasAlreadyRunning) {
+        loadUrl();
+    } else {
+        System.downloadPreloadScripts(windowIdentity, preloadScripts)
+            .catch(() => {})
```
datamadic: maybe a log? we really dont care?

src/browser/api/application.js
```diff
@@ -764,11 +761,11 @@ Application.setTrayIcon = function(identity, iconUrl, callback, errorCallback) {
     // cleanup the old one so it can be replaced
     removeTrayIcon(app);
 
-    let mainWindowIdentity = app.identity;
+    const mainWindowIdentity = app.identity;
 
     iconUrl = Window.getAbsolutePath(mainWindowIdentity, iconUrl);
 
-    cachedFetch(app.uuid, iconUrl, (error, iconFilepath) => {
+    cachedFetch(mainWindowIdentity, iconUrl, (error, iconFilepath) => {
```
datamadic: Is this a string? looks like further down its used as a key in an object

src/browser/core_state.ts
```diff
@@ -473,7 +473,32 @@ export function getWindowByUuidName(uuid: string, name: string): Shapes.OpenFinW
...
+    if (wnd && wnd.openfinWindow) {
+        return wnd.openfinWindow.browserWindow;
+    }
+}
+
+export function getWebContents(identity: Shapes.Identity): webContents|undefined {
+    const browserWindow = getBrowserWindow(identity);
+
+    if (browserWindow) {
+        return browserWindow.webContents;
```
datamadic: Do we need an `isDestroyed` check?

src/modules.d.ts
```diff
@@ -39,7 +39,7 @@ declare module 'electron' {
     }
 
     export namespace net {
-        export function request(url: string): any;
+        export function request(url: string|Object): any;
```
datamadic: Did the backing electron code change? Do we use it somewhere other than [here](https://github.com/HadoukenIO/core/pull/423/files#diff-f2ab184e212da874890dc8d44365e5bcR199)

src/browser/cached_resource_fetcher.ts
whyn07m3: Thanks for pointing this out, missed it, will add a check

src/browser/api/application.js
whyn07m3: We will catch most probable errors elsewhere. There was no logging here before, but I can add some generic error logging here no problem

src/browser/api/application.js
whyn07m3: This is not a string. Don't see where it's used as a string further down. The arguments shape of the `cachedFetch` function was changed

src/browser/api/application.js
whyn07m3: So, so far, it's hasn't affected it in any negative way I could notice while doing automated and manual testing. I do agree though it's probably better to put that event after `loadURL` just like it was before, just to keep things safe.

whyn07m3:
@datamadic Regarding manual testing, writing a test will take about 1-2 days (took 2.5 days for this PR). But that's not the main reason for not writing it. Manual test would have to have its own custom server, and to do it and add to our "Collection" of manual tests, it would have to have Node packaged in appAssets which is 30MB+, plus, a lot of glue logic.
I just think that we would have to have a framework first if we wanted to write such custom manual tests.

src/modules.d.ts
datamadic: where do we pass in an obj?

src/browser/deferred.ts
```diff
@@ -32,7 +33,7 @@ interface Bounds {
 }
 
 function handleMove(windowId: string, bounds: Bounds): void {
-    const browserWindow = BrowserWindow.fromId(windowId);
+    const browserWindow = <BrowserWindowOF>BrowserWindow.fromId(windowId);
```
nicholasdgoodman: perhaps `BrowserWindow.fromId(windowId) as BrowserWindowOF` ... this is considered the newer (better?) way of doing it. I cannot find what I read a long time ago to justify this, but perhaps we should agree on one style or the other.

src/browser/api/application.js
whyn07m3: I don't see problems with this... ¯\_(ツ)_/¯

src/browser/core_state.ts
whyn07m3: Ok, I'll add it 👍

src/browser/deferred.ts
whyn07m3: Ok, why not, will change 👍

whyn07m3:
@datamadic Not sure I fully understand your proposed solution, but I don't see how to easily and quickly add a test for this PR. Maybe we should discuss this in person.
Just to be clear, I **really-really** wanted to add a test, I love adding tests, but couldn't see an easy way to add one.

src/browser/core_state.ts
whyn07m3: [Added to `getBrowserWindow` instead](https://github.com/HadoukenIO/core/pull/423/files#diff-5990126ad71fa4d6d779051f5546f991R505)

src/browser/deferred.ts
whyn07m3: Nope, tslint gave an error, Microsoft thinks otherwise - check `prefer-type-cast` [here](https://github.com/Microsoft/tslint-microsoft-contrib)

src/browser/api/application.js
datamadic: 👍 yep, sorry

datamadic:
Cool, looks good. perhaps rerun the tests?

whyn07m3:
Test results after rebasing onto latest:
✅ [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b1154ebf2f2e22fa4b7c260)
✅ [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b11563bf2f2e22fa4b7c261)

PR #428, opened 2018-05-29
RUN-4068 - Multi runtime services and slight refactor

tgoc99:
Did a slight refactor of services to:

- make multi-runtime work
- make it so a single app could both register as a service and connect to other services as a client
- make future changes to the handshake easier

Instead of using the UUID to store/lookup the service (on both core & adapters), we now use a "channelId".  ChannelId is currently just uuid/${number} where 'number' is just an incrementing number.  This new property will future changes to the handshake and the way the channel is stored in the core / adapters a bit easier.  The adapters are also now sending back the serviceIdentity (with channelID property) along with every message after the connection.  

This change should retain backward compatibility.  

Required javacript-adapter and js-adapter changes as well.  Ran the tests with and without the adapter changes to ensure backwards compatibility.  

[Win7 - w adapter chg](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b0da955f2f2e22fa4b7c23a) 
[Win10 - w adapter chg](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b0daa95f2f2e22fa4b7c23c)

[Win7 - w/o adapter](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b0da955f2f2e22fa4b7c23b) 
[Win10 - w/o adapter](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b0daa95f2f2e22fa4b7c23d)

### DIFF
--- a/src/browser/api/service.ts
+++ b/src/browser/api/service.ts
@@ -40,35 +40,36 @@ export module Service {
     // Could be any identifier
     export function getServiceByUuid(uuid: string): ServiceIdentity {
         let serviceIdentity;
-        serviceMap.forEach((value, key) => {
-            if (value.uuid === uuid) {
-                serviceIdentity = serviceMap.get(key);
+        serviceMap.forEach(service => {
+            if (service.uuid === uuid) {
+                serviceIdentity = service;
             }
         });
         return serviceIdentity;
     }
 
-    export function registerService (identity: Identity, serviceName?: string): ServiceIdentity | false {
-        const serviceIdentity = getExternalOrOfWindowIdentity(identity);
+    export function registerService (identity: Identity, serviceName?: string): ServiceIdentity {
+        const targetApp = getExternalOrOfWindowIdentity(identity);
         // If a service is already registered from that uuid, nack
-        if (!serviceIdentity || getServiceByUuid(identity.uuid)) {
-            return false;
+        if (!targetApp || getServiceByUuid(identity.uuid)) {
+            const nackString = 'Register Failed: Please note that only one service may be registered per application.';
+            throw new Error(nackString);
         }
-        serviceIdentity.serviceName = serviceName;
-        const { uuid, isExternal, ...eventPayload } = serviceIdentity;
-        serviceIdentity.channelId = `${uuid}/${serviceId}`;
+        const { uuid, isExternal } = targetApp;
+        const channelId = `${uuid}/${serviceId}`;
+        const serviceIdentity = { ...targetApp, serviceName, channelId };
         serviceId = serviceId + 1;
 
-        serviceMap.set(serviceIdentity.channelId, serviceIdentity);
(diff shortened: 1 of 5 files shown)


nicholasdgoodman:
@tgoc99 - any reason channel ID is using a trailing number instead of a GUID? other internally generated ID's are GUIDs: (unnamed window auto-names, frame names, etc)

src/browser/api/service.ts
```diff
@@ -32,15 +33,15 @@ export module Service {
...
     }
 
     // Could be any identifier
-    export function getServiceByName(name: string): ServiceIdentity {
+    export function getServiceByUuid(uuid: string): ServiceIdentity {
         let serviceIdentity;
         serviceMap.forEach((value, key) => {
-            if (value.serviceName === name) {
+            if (value.uuid === uuid) {
                 serviceIdentity = serviceMap.get(key);
```
nicholasdgoodman: Isn't the result of this `.get(..)` call just the same as `value` to start with?

Consider:

```
serviceMap.forEach(service => {
  if (service.uuid === uuid) {
    serviceIdentity = service;
  }
});
```

or (possibly) better, standard loop instead of `forEach` to allow for short-circuit:

```
for(let serviceIdentity of serviceMap.values()) {
  if (serviceIdentity.uuid === uuid) {
    return serviceIdentity;
  }
}
```

src/browser/api/service.ts
```diff
@@ -50,18 +51,20 @@ export module Service {
     export function registerService (identity: Identity, serviceName?: string): ServiceIdentity | false {
         const serviceIdentity = getExternalOrOfWindowIdentity(identity);
         // If a service is already registered from that uuid, nack
-        if (!serviceIdentity || serviceMap.get(serviceIdentity.uuid)) {
+        if (!serviceIdentity || getServiceByUuid(identity.uuid)) {
             return false;
         }
         serviceIdentity.serviceName = serviceName;
```
nicholasdgoodman: Here I see a variable `serviceIdentity` which by convention, I would assume is of Interface type `ServiceIdentity`... (particularly because it is subsequently added to `serviceMap` of type `<string, ServiceIdentity>`.

Assigning `serviceName` should not be allowed here.

src/browser/api/service.ts
```diff
@@ -50,18 +51,20 @@ export module Service {
     export function registerService (identity: Identity, serviceName?: string): ServiceIdentity | false {
```
nicholasdgoodman: Why are we returning a union type here? This is a hack to use Javascript style "weakly typed" functions. API implementations should NAK by throwing errors, not returning "false".

tgoc99:
@nicholasdgoodman suggested changes made.  I left it just UUID/number for now as the idea was that you would be able to get the UUID from the channelID.  Honestly this is just a placeholder so that we can make channelID whatever we want to make it in the future.

PR #433, opened 2018-06-07
Feature/run 4056 global accelerators

rdepena:
### Adding support for global hotkeys. 

The API will register a given hotkey to the requesting uuid. Within the same uuid the API is asymmetrical, meaning  you can call the register method multiple times and attach several listeners to the same hotkey but a single unregister call will remove all listeners. To compensate for this fact events are raised on register/unregister.

[js-adapter sister pr](https://github.com/HadoukenIO/js-adapter/pull/135)

Tests run with core/js-adapter.asar and hadouken-js-adapter module replaced:

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b18a93af2f2e22fa4b7c2ab)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b18aa7df2f2e22fa4b7c2ac)


src/browser/api/global_hotkey.ts
```diff
@@ -0,0 +1,153 @@
...
+        if (globalShortcut.isRegistered(hotkey)) {
+            return new HotKeyError(hotkey, 'already registered');
+        }
+    }
+}
+
+export module GlobalHotkey {
+
+    export function register(identity: Identity, hotkey: string, listener: Function): void {
+        const validationError = validateRegistration(identity, hotkey);
```
pbaize: Why don't we throw directly in validateRegistration rather than not throw if we return undefined?

src/browser/api/global_hotkey.ts
```diff
@@ -0,0 +1,153 @@
...
+const subCount: Map<string, number> = new Map();
+const emitter = new EventEmitter();
+
+const eventNames = {
+    REGISTERED: 'registered',
+    UNREGISTERED: 'unregistered'
+};
+
+//index.js will use these on demand
+export const reservedHotKeys: Array<string> = [
```
pbaize: String Enum here too?

src/browser/api/global_hotkey.ts
```diff
@@ -0,0 +1,153 @@
...
+
+//require here because of missing definitions.
+const { globalShortcut } = require('electron');
+
+const subscriptionManager = new SubscriptionManager();
+const hotkeyOwnershipMap: Map<string, string> = new Map();
+const subCount: Map<string, number> = new Map();
+const emitter = new EventEmitter();
+
+const eventNames = {
```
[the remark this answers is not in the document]
rdepena: String enums are in 2.4, we are in 2.2

src/browser/api/global_hotkey.ts
rdepena: That is a good point, making the change.

PR #439, opened 2018-06-13
RUN-4042: remove process from monitor when release is called

BoOpenfin:
When release a running external process the first time, better remove it from internal [process monitor](https://github.com/openfin/runtime/blob/develop/docs/api/process-monitor.md) as well.

### DIFF
--- a/src/browser/process_tracker.js
+++ b/src/browser/process_tracker.js
@@ -328,6 +328,10 @@ ProcessTracker.prototype.release = function(uuid) {
         throw new Error(`Error releasing external process, cannot release nonpersistent processes`);
     }
 
+    if (this._processes[pid].monitor) {
+        this._processMonitor.remove(this._processes[pid].process);
+    }
+
     this._processes[pid].monitor = false;
 };
 


nicholasdgoodman:
I am not sure this should be merged.

There are three "state maps" inside this file:
- `_processes`
- `_uuidToPid`
- `_processMonitor` (via `add` and `remove` functions)

The first two of these are cleaned up by the `ProcessTracker._cleanup` function, which is called at the end of the `process-terminated` event. Meanwhile, the `ProcessMonitor` (native class) automatically removes entries from its internal map when a process terminates.

Manually calling `remove` on this class will cause the other two state objects to never be cleaned up. Additionally, as-written, in most cases you would have to call `releaseMonitor` twice to have the desired effect (most cases, the `.monitor === true` the first time).

The existing logic, by-design always keeps track of processes which have been known at some point by the application. Even on release, the core will continue to monitor the process handle until an actual termination event. This behavior is required for certain external application features.

BoOpenfin:
@nicholasdgoodman make sense to me now.
Before i just wondered why the **release** function just mark the process state as false without real deleting it from ```ProcessMonitor```.

PR #447, opened 2018-07-02
RUN-4072 window restore fix for disconnected displays

whyn07m3:
ℹ️ **About**
This is a fix for the cases when windows are being restored from their minimized position into locations outside of currently visible area because the monitor where it was located before minimizing it is now disconnected.

**Test results**
✅ [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b353fc854b21953031f3653)
✅ [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3540dc54b21953031f3654)


src/browser/api/window.js
```diff
@@ -57,6 +57,7 @@ import {
     DEFAULT_RESIZE_REGION_BOTTOM_RIGHT_CORNER,
     DEFAULT_RESIZE_SIDES
 } from '../../shapes';
+import { windowSetBoundsToVisible } from '../utils';
```
HarsimranSingh: Why is this a separate line from the previous utils import of `clipBounds`?

whyn07m3: fixed 👍

PR #448, opened 2018-07-03
RUN-4032 - Fix for Application getInfo

tgoc99:
This used to get the root ancestors manifest information, and now gets the closest parent (i.e. its own if launched from manifest, if not recursively check if parent was and return the first manifest.) . I also added initialOptions for when an app is not launched from manifest.  

I added a UUID check to [this test](https://testing-dashboard.openfin.co/#/app/tests/5b118f15f2f2e22fa4b7c266/edit) to ensure it is the correct manifest.

[PR for change to javascript-adapter docs](https://github.com/openfin/javascript-adapter/pull/447)

[Win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3b85ed54b21953031f366b)
[Win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3b875354b21953031f366e)
 [redo 1 failed win7 test](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3b886854b21953031f3670)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -343,14 +343,22 @@ Application.getShortcuts = function(identity, callback, errorCallback) {
 };
 
 Application.getInfo = function(identity, callback) {
-    const app = Application.wrap(identity.uuid);
-    const manifestObj = coreState.getManifest(identity);
-    const { url: manifestUrl, manifest } = manifestObj;
+    const app = coreState.appByUuid(identity.uuid);
+    let manifestUrl = app && app._configUrl;
+    let manifest;
+    if (manifestUrl) {
+        manifest = coreState.getManifestByUrl(manifestUrl);
+    } else {
+        const manifestObj = coreState.getManifest(identity);
+        manifestUrl = manifestObj && manifestObj.url;
+        manifest = manifestObj && manifestObj.manifest;
+    }
 
     const parentUuid = Application.getParentApplication(identity);
+    const launchMode = app && app.appObj && app.appObj.launchMode;
 
     const response = {
-        launchMode: app.launchMode,
+        launchMode,
         manifestUrl,
         manifest,
         parentUuid
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -347,8 +347,7 @@ Application.getInfo = function(identity, callback) {
     const manifestObj = coreState.getManifest(identity);
     const { url: manifestUrl, manifest } = manifestObj;
 
-    const parentApp = coreState.getAppAncestor(identity.uuid);
-    const parentUuid = (parentApp.uuid === identity.uuid) ? null : parentApp.uuid;
+    const parentUuid = Application.getParentApplication(identity);
 
     const response = {
         launchMode: app.launchMode,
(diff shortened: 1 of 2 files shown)


src/browser/api/application.js
```diff
@@ -169,14 +170,15 @@ Application.create = function(opts, configUrl = '', parentIdentity = {}) {
...
+    const app = coreState.appByUuid(opts.uuid);
 
     if (parentIdentity && parentIdentity.uuid) {
         // This is a reference to the meta `app` object that is stored in core state,
         // not the actual `application` object created above. Here we are attaching the parent
         // identity to it.
-        const app = coreState.appByUuid(opts.uuid);
         app.parentUuid = parentUuid;
     }
+    app.initialAppOptions = initialAppOptions;
```
whyn07m3: By adding this, we should also update app's shape [here](https://github.com/tgoc99/core/blob/667ea92069e9cdb69de941ecf34edcb122a65083/src/shapes.ts#L76)

src/browser/api/application.js
```diff
@@ -132,6 +132,7 @@ Application.create = function(opts, configUrl = '', parentIdentity = {}) {
 
     let appUrl = opts.url;
     const { uuid, name } = opts;
+    const initialAppOptions = Object.assign({}, opts);
```
whyn07m3: Just make sure it's okay that it is only a shallow clone

src/browser/api/application.js
whyn07m3: Also, is this different from app's `_options` ?

src/browser/api/application.js
```diff
@@ -343,16 +345,18 @@ Application.getShortcuts = function(identity, callback, errorCallback) {
...
-    const { url: manifestUrl, manifest } = manifestObj;
+    const app = coreState.appByUuid(identity.uuid);
+
+    const manifestObj = coreState.getClosestManifest(identity);
+    const { url, manifest } = manifestObj || {};
 
-    const parentApp = coreState.getAppAncestor(identity.uuid);
-    const parentUuid = (parentApp.uuid === identity.uuid) ? null : parentApp.uuid;
+    const parentUuid = Application.getParentApplication(identity);
+    const launchMode = app && app.appObj && app.appObj.launchMode;
```
whyn07m3: This was before a string or `undefined`, now can also be `false`... ```¯\_(ツ)_/¯```

src/browser/api/application.js
nicholasdgoodman: `_options` are mutated some instead of as-received. `name` is overwritten with `uuid`, `autoShow` is overwritten to false, etc.

src/browser/api/application.js
tgoc99: I believe it should only have returned a string before (at least according to docs), I think this is unnecessary protection.  The call was erroring out in the core (due to [this](https://github.com/HadoukenIO/core/pull/449)) so I added it when trying to determine why and kept it in as an unnecessary failsafe...

src/browser/api/application.js
tgoc99: :+1: we good

PR #457, opened 2018-07-06
Run 4184/event propagation

pbaize:
This centralizes event propagation from window to application and system and from application to system. I also added some unit tests for this behavior.

[win7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3eac8454b21953031f3696)
[win10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3eadd654b21953031f3697)


pbaize:
@StevenEBarbaro all the blurred decorator did is take the event and propagate it to System and Application. Since I'm now doing it from OfEvents.ts we no longer need that (or the focussedDecorator)

src/browser/of_events.ts
```diff
@@ -39,9 +41,36 @@ class OFEvents extends EventEmitter {
...
+                } else if (appWindowEventsNotOnWindow.some(t => t === topic)) {
+                    eventPropagations.set(route.system(topic), {...payload, type: propTopic, topic: 'system'});
+                }
+            }
         }
-
-        return super.emit(routeString, ...data);
+        const result = super.emit(routeString, ...data);
+        eventPropagations.forEach((payload, eventString) => {
+            this.emit(eventString, payload, ...extraArgs);
```
datamadic: Isn't this the same as `...data`?

pbaize: No, I'm changing the type and topic on the payload when I set it in the map.

src/browser/navigation_validation.ts
```diff
@@ -86,12 +86,6 @@ export function navigationValidator(uuid: string, name: string, id: number) {
                 url,
                 sourceName
             });
-            ofEvents.emit(route.application('window-navigation-rejected', uuid), {
```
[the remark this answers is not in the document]
pbaize: Line 83, and then gets propagated centrally

datamadic:
Also, why are there so many failed tests?

src/browser/navigation_validation.ts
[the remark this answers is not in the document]
datamadic: 👍

src/browser/of_events.ts
```diff
@@ -21,11 +21,13 @@ class OFEvents extends EventEmitter {
         super();
     }
 
-    public emit(routeString: string, ...data: any[]) {
+    public emit(routeString: string, payload: any, ...extraArgs: any[]) {
         const tokenizedRoute = routeString.split('/');
-
+        const eventPropagations = new Map<string, any>();
+        const data = [payload, ...extraArgs];
```
nicholasdgoodman: So now `this.emit(...)` and `super.emit(...)` have different call signatures?

Consider instead keeping the original declaration, and defining `payload` and `extraArgs` from `data` instead of vice-versa:

```
public emit(routeString: string, ...data: any[]) {
  // ...
  const [payload, ...extraArgs] = data;
  // ...
```

src/browser/of_events.ts
```diff
@@ -39,9 +41,36 @@ class OFEvents extends EventEmitter {
                 // Wildcard on any channel/topic of a specified source (ex: 'window/*/myUUID-myWindow')
                 super.emit(route(channel, '*', source), envelope);
             }
+            if (channel === 'window') {
+                const propTopic = `window-${topic}`;
+                const dontPropagateToSystem = [
+                    'auth-requested'
+                ];
+                eventPropagations.set(route.application(propTopic, uuid), {...payload, type: propTopic, topic: 'application'});
```
nicholasdgoodman: If we are going to be using the new (stage 4) spread syntax for an object literal and we are spreading on the `payload` variable, it should be declared as `object` instead of `any` because it is unlikely that we are intending to spread primitive types.

pbaize:
[Here's](https://testing-dashboard.openfin.co/#/app/tests/5b3f82bc54b21953031f3699/edit) a new test for bounds-changed at the system level

src/browser/of_events.ts
nicholasdgoodman: Suggest renaming `payload` in the `forEach` lambda to avoid confusion. Perhaps `propagationPayload` or similar...

nicholasdgoodman: Also, does the event propagation not affect the `result` return value here? Function `this.emit(...)` apparently returns something, but the value is discarded.

src/browser/of_events.ts
pbaize: In this case should we keep the payload, ...extraArgs signature?

nicholasdgoodman: Not sure I see the relationship between where we declare `payload` (signature vs scoped const) and the type we declare it as. Pretty sure you can type declare `[object, any[]]` on the destructuring assignment.

whyn07m3:
Re: "Also, why are there so many failed tests?"
@datamadic those are shape tests (shouldn't be included though)

pbaize:
New test results: [10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3fce0754b21953031f36a5) [7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b3fccfb54b21953031f36a4)

PR #469, opened 2018-07-13
Run-4184/don't propagate multiple close events

pbaize:
Fixes an issue where close was being re-propagated multiple times.

[Win 10 tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b48222054b21953031f3714)
[win 7 tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b48212954b21953031f3713)


test/global_hotkey.tests.ts
```diff
@@ -202,7 +202,7 @@ describe('GlobalHotkey', () => {
 
         GlobalHotkey.register(identity, hotkey, spy);
         //we simulate a window close.
-        ofEvents.emit(route.window('closed'), identity);
+        ofEvents.emit(route.window('closed', '*'), identity);
```
datamadic: why did this need to change (just asking)?

pbaize: The only place we were listening for `route.window('closed')` (with no subtopic/uuid/name) was the subscription manager, so I made it listen to `route.window('closed`, '*') which automatically get's re-emitted on any `route.window('closed', someId)`. That means to simulate a window close we needed to add the asterisk here to match.

src/browser/subscription_manager.ts
```diff
@@ -28,7 +28,7 @@ export default class SubscriptionManager {
     constructor() {
         this.subscriptionList = new Map();
 
-        ofEvents.on(route.window('closed'), (identity: Identity) => {
+        ofEvents.on(route.window('closed', '*'), (identity: Identity) => {
```
pbaize: @datamadic  here

test/global_hotkey.tests.ts
datamadic: also, looks like [this place](https://github.com/HadoukenIO/core/blob/develop/src/browser/external_window_event_adapter.js#L136) needs the `*` emit no? @pbaize

pbaize: I don't think so, looks like the listener for that is here https://github.com/pbaize/core/blob/b1495d1065b9f31404576ab424add869f247e80a/src/browser/api/window.js#L508

datamadic: 👍 you're right

PR #40, opened 2017-05-30
RUN-3015 Fixed context menu does not display when windows are Embedded.

nicholasdgoodman:
No API changes, internal implementation of RMB menu logic only.
Manual tests shows menu behavior unchanged for standard OpenFin windows, and fixed for EmbeddedView windows.

Test Results (took multiple runs):
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/592dd91d998bc04e20cc6c90
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/592dd940998bc04e20cc6c92

PR #470, opened 2018-07-13
RUN-4182 Add unix domain sockets as a new transport

HarsimranSingh:
We're using the npm module `unix-dgram` to create a unix domain socket transport for port discovery on POSIX systems.

Tests coming soon...


src/browser/port_discovery.ts
```diff
@@ -36,7 +37,7 @@ export class PortDiscovery extends EventEmitter {
                 log.writeToLog('info', 'Constructing the copyDataTransport window.');
                 this._transport = new WMCopyData(WINDOW_CLASS_NAME, WINDOW_CLASS_NAME);
             } else {
-                // TODO: provide a unix implementation
+                this._transport = new UnixDomainSocket(UNIX_FILE_DESCRIPTOR_PREFIX);
```
HarsimranSingh: To provide some clarification on how this works: here the core is going to create the unix domain socket transport, which means it'll create a server that is bound to `UNIX_FILE_DESCRIPTOR_PREFIX` + `Date.now()`. Other runtimes will be able to find this server by querying for all open file descriptors with this prefix.

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,84 @@
...
+        });
+
+        process.on('SIGINT', () => {
+            this.cleanUpServer();
+        });
+    }
+
+    public publish(data: any): boolean {
+        const message = new Buffer(JSON.stringify(data));
+        this.getFileDescriptors().then((fds: FileDescriptor[]) => {
```
HarsimranSingh: On `publish()`, we'll look for all open file descriptors that contain `this.fileDescriptorPrefix` and create a client on the fly to send over the data. Once sent, the client will close.

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private cleanUpServer() {
+        this.server.close();
+        unlink(this.serverName);
+    }
+
+    private getFileDescriptors(): Promise<FileDescriptor[]> {
+        return new Promise<FileDescriptor[]>((resolve) => {
+            exec(`lsof -U | grep ${this.fileDescriptorPrefix}`, (error, stdout) => {
```
HarsimranSingh: We get all open file descriptors that contain the prefix by using command line tools and then parsing the output.

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.fileDescriptorPrefix = fileDescriptorPrefix;
+        this.serverName = fileDescriptorPrefix + Date.now();
+        this.server = unixDgram.createSocket('unix_dgram', (buffer: Buffer) => {
+            this.eventEmitter.emit('message', null, buffer);
+        });
+        this.server.bind(this.serverName);
+
+        app.on('will-quit', () => {
+            this.cleanUpServer();
+        });
```
whyn07m3: can be a one-liner `app.on('will-quit', this.cleanUpServer);`?
same on line #40...?

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,84 @@
...
+                .forEach((fd: FileDescriptor) => {
+                    const client: Socket = unixDgram.createSocket('unix_dgram');
+                    client.send(message, 0, message.length, fd);
+                    client.close();
+                });
+        });
+        return true;
+    }
+
+    private cleanUpServer() {
```
whyn07m3: `private cleanUpServer(): void {`?

pbaize:
Will this require changes to Port discovery in the js-adapter for Mac and Linux?

HarsimranSingh:
@pbaize this shouldn't interfere with that.

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        process.on('SIGINT', this.cleanUpServer);
+    }
+
+    public publish(data: any): boolean {
+        const message = new Buffer(JSON.stringify(data));
+        this.getFileDescriptors().then((fds: FileDescriptor[]) => {
+            fds
+                .filter((fd: FileDescriptor) => fd !== this.serverName)
+                .forEach((fd: FileDescriptor) => {
```
anguslim: There should always be exactly one file descriptor at this point, right? If so, it's probably worth printing a log warning if there's not exactly one fd

src/browser/port_discovery.ts
```diff
@@ -1,10 +1,11 @@
 import { EventEmitter } from 'events';
-import { Base, ChromiumIPC, WMCopyData } from './transport';
+import { Base, ChromiumIPC, UnixDomainSocket, WMCopyData } from './transport';
 import * as log from './log';
 import route from '../common/route';
 import { isMeshEnabled } from './connection_manager';
 import * as coreState from './core_state';
 
+const UNIX_FILE_DESCRIPTOR_PREFIX = '/tmp/of.pd.';
```
wenjunche: should PREFIX be bit longer to avoid possible collision

src/browser/transports/unix_domain_socket.ts
```diff
@@ -0,0 +1,80 @@
...
+    on: (event: string, callback?: (...args: any[]) => void) => void;
+    send: (buffer: Buffer, offset: number, length: number, path: FileDescriptor) => void;
+}
+
+class UnixDomainSocket extends BaseTransport {
+    private fileDescriptorPrefix: string;
+    private serverName: FileDescriptor;
+    private server: Socket;
+
+    constructor(fileDescriptorPrefix: string) {
```
wenjunche: fileDescriptorPrefix should probably named fileNamePrefix.   "Descriptor" means file handle.

src/browser/transports/unix_domain_socket.ts
```diff
@@ -37,7 +37,10 @@ class UnixDomainSocket extends BaseTransport {
         });
         this.server.bind(this.serverName);
 
-        app.on('window-all-closed', this.cleanUpServer);
+        // Not 100% sure why I need to use bind() here, but when I try to call
+        // this.server.close() in this.cleanUpServer(), it throws an error
+        // saying that this.server is undefined.
```
whyn07m3: Can you check if this fixes it?
```
private cleanUpServer = () => {...
```

HarsimranSingh: This worked!

HarsimranSingh:
Updated with review comments. Ran tests on Mac test runner. Results looked good (test runner pooped out before finishing.)

PR #479, opened 2018-07-19
RUN-3607: Use userData instead of userCache in getLog{,List}

anguslim:
**Theorem.** This change:
* fixes the functions' behaviour in OS X; and
* leaves the existing correct behaviour in Windows untouched.

_Proof._ OS X is currently single-tenant, so the `userData` and `userCache`
values are modified to write to a config URL-specific folder [1]. However, when
an app is started up, the logger is redirected to a different file, and more
importantly, resets the value of `userData` [2]. Ergo, `userData` now contains
the correct directory where logs reside, rather than `userCache`. ■

[1] https://github.com/HadoukenIO/core/blob/e38be77be1afd9f236bc91249cde9bb5de200582/index.js#L726,L727
[2] https://github.com/openfin/runtime/blob/fc0c28d9e3750d7bdb771a9676c11fd110f4afe9/atom/browser/api/atom_api_app.cc#L1713

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -289,7 +289,7 @@ exports.System = {
         if (pathSafeName === name) {
             var pattern = /^debug.*\.log$/;
             if (pattern.test(pathSafeName)) {
-                fs.readFile(electronApp.getPath('userCache') + '/' + pathSafeName, {
+                fs.readFile(electronApp.getPath('userData') + '/' + pathSafeName, {
                     encoding: 'utf8'
                 }, (err, data) => {
                     if (!err) {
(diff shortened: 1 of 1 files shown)


HarsimranSingh:
Additionally, the only other place `userCache` is used is for window position caching:
https://github.com/HadoukenIO/core/blob/develop/src/browser/api/window.js#L1890
https://github.com/HadoukenIO/core/blob/develop/src/browser/api/window.js#L1915

anguslim:
Log rotating will be done in RUN-4310.

The log file is redirected in the native `App::ReopenLogfile()`, which is called from the core `rotateLogs()`, which is called when the app receives the `'ready'` message. This happens after the browser process's log file is initially opened.

I didn't touch the existing single-tenant functionality, so `userCache` still seems relevant.

HarsimranSingh:
I probably should have read the PR description first :)

PR #484, opened 2018-07-20
RUN-4311 updated dev dependencies

whyn07m3:
ℹ️ **About**
This PR makes sure all our dev dependencies are on their latest versions.

⚠️ **Exceptions:**
1. `grunt-ts`: it's still on the latest v5 (5.5.1), but updating it to the latest, which is v6 beta, breaks stuff. We can look into it later...
2. `openfin-sign`: didn't touch
3. `tslint-microsoft-contrib`: updating it to the latest only matters if we copy over and use their newer tslint.json rules file. I tried it and we got 3,9xx (yes, almost 4,000) tslint errors. Changed approximately 10+ rules, but still we were at 1,2xx errors. One day we'll fix it, but not today.

Apart from `package.json` & `npm-shrinkwrap` changes, other files had to be fixed, because of linting errors after updating.

🚥 **Automated test results**
✅ [Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b51f4a054b21953031f3767)
✅ [Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b51f56a54b21953031f3769)


pbaize:
What about using grunt-shell rather than grunt-ts like the js-adapter? https://github.com/HadoukenIO/js-adapter/blob/887de4de706c664f5775ccfee562fc810a53834b/Gruntfile.js#L38

PR #488, opened 2018-07-20
RUN-4316 application-initialized event now returns uuid

luiemilio:
Adding tests now

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -240,7 +240,7 @@ limitations under the License.
 
         // main window
         if (uuid === name) {
-            eventMap.push([`application/initialized/${uuid}`, undefined]);
+            eventMap.push([`application/initialized/${uuid}`, winIdentity]);
         }
 
         eventMap.push([`window/dom-content-loaded/${uuid}-${name}`, winIdentity]);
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -240,7 +240,7 @@ limitations under the License.
 
         // main window
         if (uuid === name) {
-            eventMap.push([`application/initialized/${uuid}`, undefined]);
+            eventMap.push([`application/initialized/${uuid}`, winIdentity]);
```
HarsimranSingh: Can you confirm that this doesn't change the payload of application's "initialized" event? It's supposed to only contain the uuid.

src/renderer/api-decorator.js
```diff
@@ -240,7 +240,8 @@ limitations under the License.
 
         // main window
         if (uuid === name) {
-            eventMap.push([`application/initialized/${uuid}`, undefined]);
+            const winUuid = { uuid: winIdentity.uuid };
+            eventMap.push([`application/initialized/${uuid}`, winUuid]);
```
whyn07m3: Can be simplified to:
```
// line #243 removed
eventMap.push([`application/initialized/${uuid}`, { uuid }]);
```

PR #490, opened 2018-07-23
RUN-4322 getApplicationGroups should only return necessary properties

licui3936:
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b563e9c54b21953031f37ba)

[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b563c5754b21953031f37b8)


licui3936:
@wenjunche , yes, before i made this change, i also see the error in some case. After changing it,  i tested it.  The issue is gone.

PR #501, opened 2018-07-27
Fix mac cache path to be more like windows

HarsimranSingh:
current macOS
`~/Library/Application Support/OpenFin/<uri encoded manifest url><optional security realm>`

proposed macOS
`~/Library/Application Support/OpenFin/cache/<md4 hash of manifest url (32 characters)>/<optional security realm>/<version>`

current Windows
`%LOCALAPPDATA%\OpenFin\cache\<optional security realm>\<version>`

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 */
 
 // built\-in modules
+let crypto = require('crypto');
 let fs = require('fs');
 let path = require('path');
 let electron = require('electron');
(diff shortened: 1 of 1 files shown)


StevenEBarbaro:
We'll need to remove the notion of the manifest URL from the cache path in the future to bring multi-tenancy back

anguslim:
As a side note, maybe we should do https://appoji.jira.com/browse/RUN-4152 in the meantime so we don't have cache path (re-)setting everywhere?

HarsimranSingh:
This needs more investigation/discussion.

PR #507, opened 2018-07-31
RUN-4332 move iframe out of experimental

datamadic:
This moves the experimental api.iframe api injection config into
startup_app.api.iframe. The idea of the outer api key is to provide
the option to manage injection to windows or apps in the future.
As is, you can now decide if iframes get the openfin api or no based
on their domain (cross origin or not). As is, this option is not
inherited.

### DIFF
--- a/src/browser/convert_options.js
+++ b/src/browser/convert_options.js
@@ -24,6 +24,11 @@ import {
 } from '../shapes';
 const TRANSPARENT_WHITE = '#0FFF'; // format #ARGB
 
+const iframeBaseSettings = {
+    'crossOriginInjection': false,
+    'sameOriginInjection': true
+};
+
 // this is the 5.0 base to be sure that we are only extending what is already expected
 function five0BaseOptions() {
     return {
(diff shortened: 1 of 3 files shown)


datamadic:
@StevenEBarbaro This is the updated one, api key added and rebased to develop 

need [this runtime pr](https://github.com/openfin/runtime/pull/1097)

[updated tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b6082a954b21953031f3878) consistent with mac pass rates.

StevenEBarbaro:
Can we leverage the new key here?

https://github.com/HadoukenIO/core/pull/507/files#diff-bb76f534ea36dcd8884c45b25ab0f7b0R247
```
 newOptions.webPreferences = {
             api: newOptions..api,
```

StevenEBarbaro:
Xavier set me straight around this question. we don't want to promote all experimental options

datamadic:
@StevenEBarbaro  Added some suggested cleaning of the experimental key

PR #518, opened 2018-08-13
Expose the Electron 'move' event

aziz512:
This PR exposes the Electron `move` event in the core API as `moving`, converting the payload into the accepted bounds format (left, top, width, height, etc).


rdepena:
🌳 until review meeting.

datamadic:
Hey @aziz512  lets reopen this once we are through some more investigation and are ready to consume

PR #522, opened 2018-08-14
RUN-4374: Write app logs to RVM before closing

anguslim:
tests, which have flaky unrelated failures:
[W7 x64](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b732b3854b21953031f397b)
[W10 x64](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b732c7a54b21953031f397c)


anguslim:
Fixed merge conflict

src/browser/rvm/utils.ts
```diff
@@ -21,7 +21,7 @@ let isFlushScheduled: boolean = false;
 let totalBytes: number = 0;
 let timer: NodeJS.Timer = null;
 
-function flushConsoleMessageQueue(): void {
+export function flushConsoleMessageQueue(): void {
```
rdepena: This is an calling an async operation `sendToRVM` but does not expose any way to wait or react once that operation is complete.

At the moment there is no guarantee that this finishes.

src/browser/api/window.js
```diff
@@ -506,6 +506,9 @@ Window.create = function(id, opts) {
             let closeEventString = route.window('close-requested', uuid, name);
             let listenerCount = ofEvents.listenerCount(closeEventString);
 
+            // Ensure that any queued app logs are sent to the RVM.
+            flushConsoleMessageQueue();
```
rdepena: Is this supposed to be Window specific or Application specific ? 

On a Multi Window application this operation will be called multiple times.

anguslim: Yeah, this makes more sense as an application-level call. I moved this to just before an application `closed` event is sent to the RVM.

src/browser/rvm/utils.ts
anguslim: I made this way jankier but it now blocks (without affecting the ignoring-async-flush behaviour for existing logs).

anguslim:
OK, so we can actually dump everything on the RVM and not care about its response. The logging library used by the RVM is good enough to have internal locking, such that incoming `application-log` and `closed` events from the runtime are handled correctly (as long as the former comes before the latter, which is true).

anguslim:
Importantly, though - we can't really test this easily. I think it might make sense holding off on this until we add app logging to the test runners, at which point we can be sure that there's no problem writing app logs during shutdown.

PR #527, opened 2018-08-21
RUN-2986: Implement getDeviceId() for non-Windows systems

anguslim:
This simply returns the SHA256 hash of the MAC address.

Note that incorrect behaviour in Windows is ignored; see RUN-4451.

### DIFF
--- a/src/browser/api/system.js
+++ b/src/browser/api/system.js
@@ -251,19 +251,15 @@ exports.System = {
         if (process.platform === 'win32') {
             return electronApp.getHostToken();
         } else {
-            try {
-                const hash = crypto.createHash('sha256');
+            const hash = crypto.createHash('sha256');
 
-                let macAddress = os.networkInterfaces().en0[0].mac;
-                if (!macAddress) {
-                    throw new Error(`MAC address (${macAddress}) not defined`);
-                }
-
-                hash.update(macAddress);
-                return hash.digest('hex');
-            } catch (e) {
-                return e;
+            const macAddress = os.networkInterfaces().en0[0].mac;
+            if (!macAddress) {
+                throw new Error(`MAC address (${macAddress}) not defined`);
             }
+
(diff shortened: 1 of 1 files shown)


src/browser/api/system.js
```diff
@@ -248,7 +248,20 @@ exports.System = {
         return hash.digest('hex');
     },
     getDeviceId: function() {
-        return electronApp.getHostToken();
+        if (process.platform === 'win32') {
+            return electronApp.getHostToken();
+        } else {
+            const hash = crypto.createHash('sha256');
+
+            let macAddress = os.networkInterfaces().en0[0].mac;
```
wenjunche: safeguard os.networkInterfaces().en0 and throw exception ?

anguslim: Done

src/browser/api/system.js
```diff
@@ -248,7 +248,23 @@ exports.System = {
         return hash.digest('hex');
     },
     getDeviceId: function() {
-        return electronApp.getHostToken();
+        if (process.platform === 'win32') {
+            return electronApp.getHostToken();
+        } else {
+            try {
```
HarsimranSingh: No need to `try...catch` . The exception gets handled correctly (Nacked) before it goes back to the javascript-adapter.

src/browser/api/system.js
```diff
@@ -248,7 +248,23 @@ exports.System = {
...
+                const hash = crypto.createHash('sha256');
+
+                let macAddress = os.networkInterfaces().en0[0].mac;
```
HarsimranSingh: `const`

src/browser/api/system.js
anguslim: Removed try/catch

src/browser/api/system.js
anguslim: Added const

PR #529, opened 2018-08-23
RUN-4378 Hash winposCache filename

licui3936:
Integration Tests:
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bca1fcecb360141a7dfd10a)
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bca20efcb360141a7dfd10b)

PS: getMachineId tests failed in every version.  Ignore them.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1970,7 +1970,6 @@ function getBoundsCacheSafeFileName(identity) {
         // If an old file name exists, replace it by a new file name
         if (fs.existsSync(oldFileName)) {
             fs.renameSync(oldFileName, newFileName);
-            log.writeToLog('info', `renamed: ${oldFileName} to: ${newFileName}`);
         }
     }
     return newFileName;
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1962,15 +1962,19 @@ function getBoundsCacheSafeFileName(identity) {
     const safeName = hash.digest('hex');
     const newFileName = path.join(userCache, windowPosCacheFolder, `${safeName}.json`);
 
-    if (!fs.existsSync(newFileName)) {
-        // current old style file name
-        const oldSafeName = new Buffer(identity.uuid + '-' + identity.name).toString('hex');
-        const oldFileName = path.join(userCache, windowPosCacheFolder, `${oldSafeName}.json`);
+    try {
+        if (!fs.existsSync(newFileName)) {
+            // current old style file name
+            const oldSafeName = new Buffer(identity.uuid + '-' + identity.name).toString('hex');
+            const oldFileName = path.join(userCache, windowPosCacheFolder, `${oldSafeName}.json`);
 
-        // If an old file name exists, replace it by a new file name
-        if (fs.existsSync(oldFileName)) {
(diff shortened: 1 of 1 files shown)


wenjunche:
With different file names, this new code will not be able to get win positions saved by previous versions of Runtime.  It should probably read from files with 'long' names, and then write to new files.

licui3936:
@wenjunche From what i see,  each runtime has its own cache path.  So new version will never use older version cache file. Am i wrong?

HarsimranSingh:
@licui3936 the RVM now migrates the `winposcache` by default when an application updates its runtime version. So older style filenames can still exist in newer runtimes. We need to provide a migration path for this.

licui3936:
@HarsimranSingh @wenjunche i updated the codes. Please review it. Thanks!!

src/browser/api/window.js
```diff
@@ -1948,13 +1949,30 @@ function saveBoundsToDisk(identity, bounds, zoomLevel, callback) {
...
-    return path.join(userCache, windowPosCacheFolder, `${safeName}.json`);
+
+    // new style file name
+    const hash = crypto.createHash('sha256');
+    hash.update(identity.uuid);
+    hash.update(identity.name);
+    const safeName = hash.digest('hex');
+    const newFileName = path.join(userCache, windowPosCacheFolder, `${safeName}.json`);
+
+    if (!fs.existsSync(newFileName)) {
```
datamadic: Just a thought, if we know that this does not exist we could let the caller know, as often it will immediately try to read it (see [here](https://github.com/HadoukenIO/core/blob/46e136b0a9c052461e2fa4859b29e9a1b254eb41/src/browser/api/window.js#L1766)). Though I think this is fine as is.

licui3936: yeah, i thought that too. But i found this  function 'getBoundsCacheSafeFileName" is also called in saveBoundsToDisk  to get the cached file name.  If i want to return something not "cached file name" when it doesn't exist, i need to add something else to differentiate  these two cases.

licui3936:
@HarsimranSingh i thought about it. The reason i chose sync way is because of these reasons:  1. This is non-parallel task. It's not like something can run in the background while doing something else. The next call rely on it.  2. async function 'exists' already deprecated. Better not use it.  3. I think this operation is small so sync variety of functions are faster than async ones.

rdepena:
🌳 bumping to .36. Needs more eyes because this is window creation.

nicholasdgoodman:
@tgoc99 : just a little technicality, while `fs.exists` _is_ deprecated, `fs.existsSync` is not. probably still correct that it should not be used in this situation.

regardless, doing multiple file operations will _always_ introduce race conditions. (e.g. check for existence, and then move) this code will have the possibility of failing at each step.

if sync functions are used, i believe `try` / `catch` should be as well. if async, then `err` callbacks should be provided.

licui3936:
Add try/catch

licui3936:
@tgoc99 and @pbaize : i changed it to use async calls.

licui3936:
@HarsimranSingh i ran all the tests and updated the tests links.

PR #538, opened 2018-09-05
RUN-4519 API breadcrumbs

StevenEBarbaro:
When breadcrumbs are enabled, the debug.log contains tracking of the API messages at each hop of the system.

This is a Window option that defaults to _false_
_experimental.api.breadcrumbs_

🌳 Requires this [adapter PR 493](https://github.com/openfin/javascript-adapter/pull/493)

### DIFF
--- a/src/browser/api_protocol/transport_strategy/ack.ts
+++ b/src/browser/api_protocol/transport_strategy/ack.ts
@@ -3,6 +3,7 @@ const errors = require('../../../common/errors');
 
 export class AckMessage {
     public readonly action: string = 'ack';
+    public breadcrumbs: Array<any>;
     public correlationId: number;
     public payload: AckPayload | NackPayload;
 }
--- a/src/browser/api_protocol/transport_strategy/elipc_strategy.ts
+++ b/src/browser/api_protocol/transport_strategy/elipc_strategy.ts
@@ -99,7 +99,7 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
 
         try {
             const data = JSON.parse(JSON.stringify(rawData));
-            const ack = !data.isSync ? this.ackDecorator(e, data.messageId) : this.ackDecoratorSync(e, data.messageId);
+            const ack = !data.isSync ? this.ackDecorator(e, data.messageId, data) : this.ackDecoratorSync(e, data.messageId);
             const nack = this.nackDecorator(ack);
             const browserWindow = e.sender.getOwnerBrowserWindow();
             const currWindow = browserWindow ? coreState.getWinById(browserWindow.id) : null;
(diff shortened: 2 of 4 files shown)


src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -151,10 +151,30 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
         };
     }
 
-    protected ackDecorator(e: any, messageId: number): AckFunc {
+    private addBreadcrumb(message: any,  messageId: number, action: string, name: string): void {
+        if (message && message.breadcrumbs && message.breadcrumbs.length) {
```
anguslim: This conditional should always be true if the function is getting called, right? If so, it's probably worth putting an else block that prints an error saying the message was invalid (and the message itself).

StevenEBarbaro: This is a high-transactional call when enabled. The debug.log could become flooded with the error message. I can adjust the implementation to add the array when it is not present.

anguslim: That sounds good - maybe add the array if not present, and log?

StevenEBarbaro: I'm advocating to avoid logging, this would be called on *every* API message when enabled

anguslim: I'm thinking that because every call of `addBreadcrumb` (at the moment) is conditional on `usingBreadcrumbs` being true, we should always expect the `breadcrumbs` array to be present, so if it's not it's something we'd want to note by logging. If we expect to call `addBreadcrumb` without that conditional, then it wouldn't make as much sense to log.

rdepena:
🌳 move to .36

src/browser/api_protocol/transport_strategy/ack.ts
```diff
@@ -3,6 +3,7 @@ const errors = require('../../../common/errors');
 
 export class AckMessage {
     public readonly action: string = 'ack';
+    public breadcrumbs: Array<any>;
```
HarsimranSingh: This should be marked optional with `?`

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -151,10 +151,30 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
+        }
+    }
+
+    protected ackDecorator(e: any, messageId: number, originalPayload: any): AckFunc {
         const ackObj = new AckMessage();
         ackObj.correlationId = messageId;
 
+        // It is inferred that breadcrumbs have been enabled for a window
+        // when the breadcrumb array is present in the original payload of the API request.
+        const usingBreadcrumbs = originalPayload.breadcrumbs && originalPayload.breadcrumbs.length;
```
HarsimranSingh: Why do we need to check the length? Shouldn't it be enough that breadcrumbs exists?

StevenEBarbaro: That's to perform a sanity check against a breadcrumb enabled message originating from an adapter VS a breadcrumb formed envelope as generated in the core

PR #539, opened 2018-09-07
RUN-4368 id undefined issue

licui3936:
This PR audited application APIs and fixed a few undefined issues found when an app doesn't exist. 
Tests: 
[Windows 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b92899a6107b45e6d1ebc40)
[Windows 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5b928b016107b45e6d1ebc41)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -278,10 +278,13 @@ Application.close = function(identity, force, callback) {
     }
 };
 
-Application.getChildWindows = function(identity /*, callback, errorCallback*/ ) {
-    var app = Application.wrap(identity.uuid);
-
-    return coreState.getChildrenByApp(app.id);
+Application.getChildWindows = function(identity, callback, errorCallback) {
+    const app = Application.wrap(identity.uuid);
+    if (!app) {
+        errorCallback(new Error(`application with uuid '${identity.uuid}' does not exist`));
+    } else {
+        callback(coreState.getChildrenByApp(app.id));
+    }
 };
 
 Application.getGroups = function( /* callback, errorCallback*/ ) {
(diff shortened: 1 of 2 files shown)


rdepena:
@licui3936 What about errors on applications that are not running ?

licui3936:
@rdepena , i tested 'app not running' case,  there is no exception.  the APIs still return the results.

src/browser/api/application.js
```diff
@@ -837,18 +861,27 @@ Application.setTrayIcon = function(identity, iconUrl, callback, errorCallback) {
...
-    // set zoom level for each child window
-    app.children.forEach(function(childWindow) {
-        const childWindowIdentity = {
-            name: childWindow.openfinWindow.name,
-            uuid: childWindow.openfinWindow.uuid
-        };
-        Window.setZoomLevel(childWindowIdentity, level);
-    });
+    if (appError) {
+        errorCallback(new Error(appError));
```
HarsimranSingh: Need to check `typeof errorCallback === 'function'` just like `callback`

src/browser/api/application.js
```diff
@@ -837,18 +861,27 @@ Application.setTrayIcon = function(identity, iconUrl, callback, errorCallback) {
     });
 };
 
-Application.setZoomLevel = function(identity, level) {
-    const app = coreState.appByUuid(identity.uuid);
+Application.setZoomLevel = function(identity, level, callback, errorCallback) {
```
HarsimranSingh: https://github.com/HadoukenIO/core/blob/develop/src/browser/api_protocol/api_handlers/application.js#L94 callback and errorcallback aren't even passed in here

src/browser/api/application.js
licui3936: Fixed it. Actually we don't need callback and errorcallback in this function.

src/browser/api/application.js
licui3936: Please see other comments.

PR #553, opened 2018-09-17
RUN-4510 Add flash version which is necessary for enable-chromium build

TomaszMa:


index.js
```diff
@@ -415,6 +415,9 @@ function includeFlashPlugin() {
 
     if (pluginName) {
         app.commandLine.appendSwitch('ppapi-flash-path', path.join(process.resourcesPath, 'plugins', 'flash', pluginName));
+        // Currently for enable_chromium build the flash version need to be
+        // specified. See RUN-4510 and RUN-4580.
+        app.commandLine.appendSwitch('ppapi-flash-version', '3.0.0');
```
anguslim: Why not 30.0.0.154, the actual version?

PR #557, opened 2018-09-21
RUN-4406 - Experimental opt-in renderer batching

StevenEBarbaro:
Enables batching API messages originating from a renderer process

Test Result: https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5ba5486bd6d66a0ddf8e7287

Relates to this [Adapter PR]()

### DIFF
--- a/src/browser/api_protocol/transport_strategy/ack.ts
+++ b/src/browser/api_protocol/transport_strategy/ack.ts
@@ -5,7 +5,29 @@ export class AckMessage {
     public readonly action: string = 'ack';
     public breadcrumbs?: Array<any>;
     public correlationId: number;
+    public readonly originalAction?: string;
     public payload: AckPayload | NackPayload;
+
+    constructor(breadcrumbs?: Array<any>, originalAction?: string) {
+        if (breadcrumbs) {
+            this.breadcrumbs = breadcrumbs;
+        }
+
+        if (originalAction) {
+            this.originalAction = originalAction;
+        }
+    }
+
+    public addBreadcrumb(name: string, time?: number, messageId?: number): void {
+        this.breadcrumbs = this.breadcrumbs || [];
+
+        this.breadcrumbs.push({
+            action: this.originalAction,
+            messageId: messageId || this.correlationId,
+            name,
+            time: time || Date.now()
+        });
+    }
 }
 
 // ToDo following duplicated in src/shapes.ts
--- a/src/browser/api_protocol/transport_strategy/api_transport_base.ts
+++ b/src/browser/api_protocol/transport_strategy/api_transport_base.ts
@@ -18,6 +18,12 @@ export interface MessagePackage {
     strategyName: any; // ws / elipc
 }
 
+// To Do: Refactor api_transport_base.ts and construct implementations per type, per endpoint.
+//        Configuration can then be stored per instance.
+/* tslint:disable:no-empty-interface*/
+export interface MessageConfiguration {}
+/* tslint:enable:no-empty-interface*/
+
 export abstract class ApiTransportBase<T> {
 
     protected requestHandler: RequestHandler<T>;
@@ -36,9 +42,9 @@ export abstract class ApiTransportBase<T> {
 
     public abstract onClientDisconnect(cb: Function): void;
 
-    protected abstract onMessage(id: number, data: any): void;
+    protected abstract onMessage(id: number, data: any, ackDelegate: any): void;
 
-    protected abstract ackDecorator(id: number, messageId: number, originalPayload: any): AckFunc;
+    protected abstract ackDecorator(id: number, messageId: number, originalPayload: any, configuration: MessageConfiguration): AckFunc;
 
     protected abstract ackDecoratorSync(e: any, messageId: number): AckFunc;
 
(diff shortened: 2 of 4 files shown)


src/browser/api_protocol/transport_strategy/ack.ts
```diff
@@ -5,7 +5,29 @@ export class AckMessage {
     public readonly action: string = 'ack';
     public breadcrumbs?: Array<any>;
     public correlationId: number;
+    public readonly originalAction?: string;
     public payload: AckPayload | NackPayload;
+
+    constructor(breadcrumbs?: Array<any>, originalAction?: string) {
```
whyn07m3: Since both parameters are optional, it would probably be best to pass them in an object
Same with `addBreadcrumb` method

src/browser/api_protocol/transport_strategy/ack.ts
```diff
@@ -5,7 +5,29 @@ export class AckMessage {
...
+        if (breadcrumbs) {
+            this.breadcrumbs = breadcrumbs;
```
whyn07m3: Maybe better to initiate with given breadcrumbs or empty array, so that `addBreadcrumb` doesn't have to keep checking and can simply `push`

src/browser/api_protocol/transport_strategy/ack.ts
```diff
@@ -5,7 +5,29 @@ export class AckMessage {
...
+            this.originalAction = originalAction;
+        }
+    }
+
+    public addBreadcrumb(name: string, time?: number, messageId?: number): void {
+        this.breadcrumbs = this.breadcrumbs || [];
+
+        this.breadcrumbs.push({
+            action: this.originalAction,
+            messageId: messageId || this.correlationId,
```
whyn07m3: Can message id be 0, and if yes, is this still okay?

src/browser/api_protocol/transport_strategy/api_transport_base.ts
```diff
@@ -18,6 +18,12 @@ export interface MessagePackage {
     strategyName: any; // ws / elipc
 }
 
+// To Do: Refactor api_transport_base.ts and construct implementations per type, per endpoint.
```
whyn07m3: Just a reminder to add a ticket.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -124,10 +181,11 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
             }
 
             const identity = {
+                batch: data.action === 'api-batch',
+                entityType,
                 name: subFrameName,
-                uuid: opts.uuid,
                 parentFrame: opts.name,
-                entityType
+                uuid: opts.uuid
```
whyn07m3: :neckbeard: `opts.uuid` can be destructed on line 174

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -168,28 +235,23 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
+                            new AckMessage(originalPayload.breadcrumb, originalPayload.action));
         ackObj.correlationId = messageId;
+        return ackObj;
+    }
 
-        // It is inferred that breadcrumbs have been enabled for a window
-        // when the breadcrumb array is present in the original payload of the API request.
-        const usingBreadcrumbs = originalPayload.breadcrumbs && originalPayload.breadcrumbs.length;
+    protected ackDecorator(e: any, messageId: number, originalPayload: any, baseConfiguration: MessageConfiguration): AckFunc {
+        const configuration: ElIPCConfiguration = <ElIPCConfiguration>baseConfiguration;
```
whyn07m3: I'm confused, casting to `ElIPCConfiguration`, can't it be defined in the params as `ElIPCConfiguration` instead of `MessageConfiguration`?

Same question for line 281

src/browser/api_protocol/transport_strategy/ack.ts
StevenEBarbaro: The intent was to _only_ add the new array to every payload when breadcrumbs are present.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
```diff
@@ -205,12 +267,51 @@ export class ElipcStrategy extends ApiTransportBase<MessagePackage> {
...
+
+                if (usingBreadcrumbs) {
+                    ackObj.addBreadcrumb('core/deferredACK');
+                }
+
+                deferredAcks.push(ackObj);
+
+                if (deferredAcks.length === configuration.rendererBatchConfiguration.length) {
+                   mainAck({
+                    success: true,
```
anguslim: What happens if there's a nack? Should this still return success = true? Can the NackPayload still have breadcrumbs added to it?

StevenEBarbaro: NACK decorators invoke the `ackDecorator`
https://github.com/HadoukenIO/core/blob/00eeebfee7b993c229ae1554c37ffbdfdf4f4e6f/src/browser/api_protocol/transport_strategy/api_transport_base.ts#L47

This is a batch, not a transaction. Success implies the batch was dispatched and consumed by the core.

src/browser/api_protocol/transport_strategy/elipc_strategy.ts
StevenEBarbaro: This was recommended by @pbaize to avoid swaths of refactoring to the entire transport layer. ApiTransportBase<T> is providing the method signature and further coupling the configurations at the base-class for el_ipc specific fields was avoided.

PR #567, opened 2018-09-29
RUN-4654: calling app.exit during delegated launch (enable_chromium)

wenjunche:
Tests

[W7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5baed8a4cb360141a7dfcfc1)
[W10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5baeda1ecb360141a7dfcfc2)

### DIFF
--- a/index.js
+++ b/index.js
@@ -226,7 +226,7 @@ app.on('chrome-browser-process-created', function() {
 
         app.commandLine.appendArgument('noerrdialogs');
         process.argv.push('--noerrdialogs');
-        app.quit();
+        app.exit(0);
 
         return;
     }


index.js
```diff
@@ -226,7 +226,7 @@ app.on('chrome-browser-process-created', function() {
 
         app.commandLine.appendArgument('noerrdialogs');
         process.argv.push('--noerrdialogs');
-        app.quit();
+        app.exit(0);
```
HarsimranSingh: Why are we switching from `quit` to `exit`? `quit` seems to be used all over the core. Do all those instances need to be changed as well?

wenjunche: Added the following comments in JIRA:

The exception happens because Browser::quit tries to shut services not initialized yet for enable_chromium when 2nd instance of Runtime exits right after start-up.  Browser::exit just exits.

No need to change other calls to quit.

wenjunche:
The fix is fixed in latest develop at native side

PR #568, opened 2018-10-01
RUN-4525: refine grouped window move on MAC

BoOpenfin:
1: for full screen window, set to normal mode first.
2: No need to restore cached bounds on mac cause we hasn't [store the window's bounds before max or fullscreen](https://github.com/openfin/runtime/blob/d8f3a2aa275bf085d12716f89ea86573dd7d0bba/atom/browser/native_window_views.h#L393).

### DIFF
--- a/src/browser/bounds_changed_state_tracker.ts
+++ b/src/browser/bounds_changed_state_tracker.ts
@@ -452,11 +452,14 @@ export default class BoundsChangedStateTracker {
                             const [w, h] = [width, height];
                             wt.setWindowPos(hwnd, { x, y, w, h, flags });
                         } else {
-                            if (win.browserWindow.isMaximized()) {
+                            if (win.browserWindow.isFullScreen()) {
+                                win.browserWindow.setFullScreen(false);
+                            } else if (win.browserWindow.isMaximized()) {
                                 win.browserWindow.unmaximize();
+                            } else {
+                                // no need to call clipBounds here because called earlier
+                                win.browserWindow.setBounds({ x, y, width, height });
                             }
-                            // no need to call clipBounds here because called earlier
-                            win.browserWindow.setBounds({ x, y, width, height });
                         }
                     });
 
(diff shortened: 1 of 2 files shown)


src/browser/bounds_changed_state_tracker.ts
```diff
@@ -452,11 +452,14 @@ export default class BoundsChangedStateTracker {
                             const [w, h] = [width, height];
                             wt.setWindowPos(hwnd, { x, y, w, h, flags });
                         } else {
-                            if (win.browserWindow.isMaximized()) {
+                            if (win.browserWindow.isFullScreen()) {
+                                win.browserWindow.setFullScreen(false);
+                            } else if (win.browserWindow.isMaximized()) {
                                 win.browserWindow.unmaximize();
+                            } else {
```
HarsimranSingh: I don't think this should be in an `else` block. We want to unFullscreen/unMaximize a window before we set the bounds, correct?

BoOpenfin: Actually, cause we are using [current bounds](https://github.com/HadoukenIO/core/blob/develop/src/browser/bounds_changed_state_tracker.ts#L423) as the before bounds.
And for a fullscreen/maximize window on MAC, the bounds is the real size (not the size [before maximize](https://github.com/openfin/runtime/blob/d8f3a2aa275bf085d12716f89ea86573dd7d0bba/atom/browser/native_window_views.cc#L576) which is [the case](https://github.com/openfin/runtime/blob/d8f3a2aa275bf085d12716f89ea86573dd7d0bba/atom/browser/native_window_views.cc#L732) on WIN). 
Another option is add restore_bounds_ on the runtime side for mac just like win_os does.

src/modules.d.ts
```diff
@@ -100,8 +100,10 @@ declare module 'electron' {
         bringToFront(): any;
         isDestroyed(): boolean;
         isMaximized(): boolean;
+        isFullScreen(): boolean;
         isMinimized(): boolean;
         unmaximize(): any;
+        setFullScreen(fullscreen: boolean): any;
```
whyn07m3: this probably returns `void`

PR #571, opened 2018-10-02
RUN-4659: Replace did-get-response-details with did-frame-navigate

anguslim:
The old event is no longer emitted in Electron 3, so we can't use it to get
an HTTP response code.

This fixes at least one failing v10 integration test, and completes the
task in RUN-4448.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -121,9 +121,9 @@ let browserWindowEventMap = {
 };
 
 let webContentsEventMap = {
-    'did-get-response-details': {
-        topic: 'resource-response-received',
-        decorator: responseReceivedDecorator
+    'did-frame-navigate': {
+        topic: 'frame-navigated',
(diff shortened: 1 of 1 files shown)


src/browser/api/window.js
```diff
@@ -121,9 +121,9 @@ let browserWindowEventMap = {
 };
 
 let webContentsEventMap = {
-    'did-get-response-details': {
-        topic: 'resource-response-received',
```
HarsimranSingh: `resource-response-received` is actually is an outward facing event that has a specific shape and is documented [here](http://cdn.openfin.co/jsdocs/stable/tutorial-window.addEventListener.html). A lot of this shouldn't be changed.

anguslim:
The best course of action is to probably keep the core routing as is and fill in as many fields as we can in the native layer while emitting the same event. I'll close this

PR #588, opened 2018-10-18
RUN-4709 set the app user model id 

datamadic:
RUN-4709 set the app user model id  to `company[.name]` if exists else… `uuid[.name]`. Name is optional for both.

Tests well:
[win 10](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bc8e6d2cb360141a7dfd0fc)
[win 7](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bc8e53dcb360141a7dfd0fa)

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,6 @@ let globalShortcut = electron.globalShortcut;
 let ipc = electron.ipcMain;
 let Menu = electron.Menu;
 
-
 // npm modules
 let _ = require('underscore');
 let minimist = require('minimist');
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@ let globalShortcut = electron.globalShortcut;
 let ipc = electron.ipcMain;
 let Menu = electron.Menu;
 
+
 // npm modules
 let _ = require('underscore');
 let minimist = require('minimist');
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -560,10 +560,25 @@ function launchApp(argo, startExternalAdapterServer) {
...
+
+        if (company) {
+            namePart = shortcutName ? `.${shortcutName}` : '';
+            appUserModelId = `${company}${namePart}`;
+        } else {
+            namePart = name ? `.${name}` : '';
+            appUserModelId = `${uuid}${namePart}`;
+        }
+
+        app.setAppUserModelId(appUserModelId);
```
HarsimranSingh: `BrowserWindow.setTaskbarGroup()` does some stuff with AppUserModelId. Can we make sure that these things play well together?

https://github.com/openfin/runtime/blob/develop/atom/browser/native_window_views_win.cc#L735

wenjunche: To follow up to Sim's comment,  I did some quick tests and here are the findings:  Electron actually sets a default model Id with "electron.app.OpenFin", which explains why I saw notifications in a group called "electron.app.OpenFin" on Window server 2012.  So, we should set it to what we want.

app.setAppUserModelId does not seem to affect taskbar group code (link from Sim's comment)

PR #600, opened 2018-10-25
 RUN-4690 only resize inside threshold plus move rectangle math 

datamadic:
There are a few things that this PR does, but mostly it is a precursor to making group resizing a deferred-bounds transaction by starting to tease out and isolate the logic that handles the positioning from the logic that decides how and when to move it. 

The three changes to behavior that this makes are:
*  Windows will now only resize if they are within 5px of each other. Currently, for example, a vertical resize will resize any window resting on that `x` value no matter where it is on the screen
* Windows now, when they are within the 5 px threshold and are a candidate for a grouped resize will take the value of the resizing edge exactly, meaning that they snap to the resizing edge and stick there.
* When a window that is being resized reaches its minimum size it will be pushed in the direction of the resize instead of losing the connected edge. 

Also, hopefully you find that the changes here are generally more stable than whats currently there. In testing / investigating it seems like there is quite a bit of perhaps unintentional behavior associated with how it works now.

We need to validate these changes with layouts to make sure that we don't break them in a bad way. This should be treated as an intermediate step towards a larger behavior change, though should be consumable and stable as is.  

[Tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bd32ef0cb360141a7dfd18d)

### DIFF
--- a/src/browser/bounds_changed_state_tracker.ts
+++ b/src/browser/bounds_changed_state_tracker.ts
@@ -427,12 +427,24 @@ export default class BoundsChangedStateTracker {
                                 win.browserWindow.unmaximize();
                             }
 
-                            setPositions.push(() => {
-                                const myBounds = positions.get(win.name);
-                                const { x, y, width, height } = myBounds;
-                                const [w, h] = [width, height];
-                                wt.setWindowPos(hwnd, { x, y, w, h, flags });
-                            });
+                            /*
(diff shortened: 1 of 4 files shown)


datamadic:
@rdepena @pbaize lets treat this a the new normal. Any bugs / regressions are real and we need to address pre 37 shipping. Game on

PR #602, opened 2018-10-27
RUN-4480: no API injection for Chrome pages

wenjunche:
[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bd3b961cb360141a7dfd19b)
[Runtime](https://github.com/openfin/runtime/pull/1224)

Chrome pages need the original window.alert, confirm, prompt.

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -1102,7 +1102,7 @@ function createAppObj(uuid, opts, configUrl = '') {
         // save the original value of autoShow, but set it false so we can
         // show only after the DOMContentLoaded event to prevent the flash
         opts.toShowOnRun = eOpts['autoShow'];
-        eOpts.show = false;
+        eOpts.show = isChromePageUrl(opts.url); // no API injection for chrome pages
 
         appObj.mainWindow = new BrowserWindow(eOpts);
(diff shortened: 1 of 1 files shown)


src/browser/api/application.js
```diff
@@ -1102,7 +1102,7 @@ function createAppObj(uuid, opts, configUrl = '') {
         // save the original value of autoShow, but set it false so we can
         // show only after the DOMContentLoaded event to prevent the flash
         opts.toShowOnRun = eOpts['autoShow'];
-        eOpts.show = false;
+        eOpts.show = isChromePageUrl(opts.url); // no API injection for chrome pages
```
HarsimranSingh: Maybe this way would show a clearer intent?
```js
// save the original value of autoShow, but set it false so we can
// show only after the DOMContentLoaded event to prevent the flash.
// don't do this for chrome pages since the openfin api isn't injected
if (!isChromePageUrl(opts.url)) {
  opts.toShowOnRun = eOpts['autoShow'];
  eOpts.show = false;
}
```

wenjunche: updated

PR #609, opened 2018-11-05
RUN-4734: better solution for disabling API injection for chrome page…

wenjunche:
Instead of checking if API should be injected in init.js,  it is done in RenderClient::DidCreateScriptContext

[Runtime PR](https://github.com/openfin/runtime/pull/1238)

[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5be0a076cb360141a7dfd1fd)


wenjunche:
Added RUN-4817 for listener cleaning up

PR #618, opened 2018-11-08
RUN-4497 restrict group moves that break bounds

datamadic:


src/browser/disabled_frame_group_tracker.ts
```diff
@@ -176,3 +198,30 @@ export function removeWindowFromGroup(win: OpenFinWindow) {
...
     groupInfoCache.delete(groupUuid);
 }
+
+interface Clamped {
+    value: number;
+    clampedOffset: number;
+}
+
+function clipBounds(bounds: Rectangle, browserWindow: OpenFinWindow['browserWindow']): Rectangle {
+    if (!('_options' in browserWindow)) {
```
rdepena: When do you see `_options` missing from the `browserWindow` objects ?

PR #619, opened 2018-11-08
RUN-4377 Handle new event "bounds-changing"

TomaszMa:
On Windows with performance mod enabled event "bounds-changing" is not emitted during user drag. I added new event to be emitted from runtime and created commit with support for this new runtime event.
This should be integrated together with https://github.com/openfin/runtime/pull/1240

### DIFF
--- a/src/browser/bounds_changed_state_tracker.ts
+++ b/src/browser/bounds_changed_state_tracker.ts
@@ -98,6 +98,9 @@ export default class BoundsChangedStateTracker {
                 ofEvents.emit(route.window('end-user-bounds-changing', uuid, name), Object.assign(payload, bounds));
                 this.handleBoundsChange(false, true);
             },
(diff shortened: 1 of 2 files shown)


src/browser/bounds_changed_state_tracker.ts
```diff
@@ -281,11 +284,29 @@ export default class BoundsChangedStateTracker {
...
+            if (this.browserWindow.isMinimized()) {
+                windowState = WindowState.Minimized;
+            }
+            const frame = this.browserWindow._options.frame;
+
+            currentBounds = {
+              ...bounds,
+              frame,
+              windowState
+            };
```
wenjunche: instead of duplicating this.getCurrentBounds here, easier to call this.getCurrentBounds() and override (x,y) from bounds ?

PR #642, opened 2018-11-20
RUN-4152 migrate cache path from runtime to core

licui3936:
This PR should combine with [runtime PR](https://github.com/openfin/runtime/pull/1255) together.

[Tests](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5bfc61c1cb360141a7dfd3b3)


HarsimranSingh:
After further discussion with @licui3936 and @anguslim, moving the cachepath logic to the core isn't the best idea. For logging of all processes (browser, renderer, utility) to work correctly, the cachepath needs to be set as early as possible.

PR #655, opened 2018-11-27
RUN-4690 Formalize grouped resize behavior

datamadic:
Only resize the windows in the group when they are closer than the bounds share threshold

## Hold till merge back to develop
I will rebase after merge, but this is the change set


datamadic:
see rebase here https://github.com/HadoukenIO/core/pull/661

PR #656, opened 2018-11-28
Merge staging into develop

pbaize:


pbaize:
This may need a follow up PR to update the js-adapter in develop

HarsimranSingh:
@pbaize the js-adapter will update automatically now. No need to manually set up a PR.

PR #661, opened 2018-12-03
RUN-4690 Formalize grouped resize behavior

datamadic:
When in a group, a resizing window will only grab the edeges of
windows that are within the shared bounds threshold to the moving
edge

@HarsimranSingh

### DIFF
--- a/src/browser/rectangle.ts
+++ b/src/browser/rectangle.ts
@@ -373,6 +373,13 @@ export class Rectangle {
         return Array.from(Rectangle.ADJACENCY_LIST([...rects, this as Rectangle]).values()).find(list => list.includes(this));
     }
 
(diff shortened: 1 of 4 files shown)


src/browser/rectangle.ts
```diff
@@ -373,6 +373,13 @@ export class Rectangle {
         return Array.from(Rectangle.ADJACENCY_LIST([...rects, this as Rectangle]).values()).find(list => list.includes(this));
     }
 
+    public hasIdenticalBounds(rect: RectangleBase): boolean {
```
rdepena: function is using `this`, let's move this to an `=>` declaration.

datamadic:
to be followed up with a more different PR

PR #669, opened 2018-12-06
RUN-4794 Fix js exception popup issue

licui3936:
This PR is to fix the js exception issue when calling system.clearCache during app startup.  
[Windows 10 Test](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5c097759cb360141a7dfd594)

### DIFF
--- a/src/browser/cached_resource_fetcher.ts
+++ b/src/browser/cached_resource_fetcher.ts
@@ -9,8 +9,8 @@ import { addPendingAuthRequests, createAuthUI } from './authentication_delegate'
 import { AuthCallback, Identity } from '../shapes';
 import { getSession } from './core_state';
 
-let appQuiting: Boolean = false;
-let clearCache: boolean = false;
+let appQuiting: boolean = false;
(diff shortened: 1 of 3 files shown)


src/browser/cached_resource_fetcher.ts
```diff
@@ -10,6 +10,7 @@ import { AuthCallback, Identity } from '../shapes';
 import { getSession } from './core_state';
 
 let appQuiting: Boolean = false;
+let clearCache: boolean = false;
```
HarsimranSingh: :neckbeard: Do both `Boolean` and `boolean` work in Typescript?

nicholasdgoodman: both work, but the upper-cased `Boolean` should never be used:
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

licui3936:
@datamadic and @HarsimranSingh:
i just updated the codes based on your reviews.

PR #672, opened 2018-12-06
RUN-4834 Updated Ack to send logId in a data payload.

arjun2k15:
The current implementation just receives an Ack or a Nack depending on whether the log was uploaded. Now we pass along a payload with the logId for the log that was uploaded.

To run:
```
app = fin.desktop.Application.getCurrent();

app.sendApplicationLog((info) => {
  console.log('logId', info.logId);
}, (err) => {
  console.log(err);
});
```

### DIFF
--- a/src/browser/api_protocol/api_handlers/application.js
+++ b/src/browser/api_protocol/api_handlers/application.js
@@ -82,7 +82,11 @@ function sendApplicationLog(identity, message, ack) {
     const payload = message.payload;
     const appIdentity = apiProtocolBase.getTargetApplicationIdentity(payload);
 
-    return Application.sendApplicationLog(appIdentity).then(() => ack(successAck));
+    return Application.sendApplicationLog(appIdentity).then((logId) => {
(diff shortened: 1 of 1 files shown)


arjun2k15:
There is a chance it could fail - incorrect license credentials, wrong url, etc. The Rvm sends a Nack depending on the log manager response in the form:
```
{
  success: false,
  messageId: strMessageId,
  error: strErrorMsg
}
```

arjun2k15:
I believe that the error will Nack in elipc_strategy.js. It seems to catch the rvm json message and nack it back to the adapter's errorCallback function.

rdepena:
ah, perfect  👍

HarsimranSingh:
Link to the javascript-adapter PR: https://github.com/openfin/javascript-adapter/pull/515

@arjun2k15 we need a js-adapter PR as well, correct?

arjun2k15:
Just made a PR in the JS adapter.

https://github.com/HadoukenIO/js-adapter/pull/217

PR #677, opened 2018-12-12
Windows with the same name can now join a group.

rdepena:
Initially was using nativeId but windows leave groups during cleanup so the browserWindow has a change of being destroyed.

Auto Tests running.
Manual Tests pending.

### DIFF
--- a/src/browser/window_groups.ts
+++ b/src/browser/window_groups.ts
@@ -72,6 +72,11 @@ export class WindowGroups extends EventEmitter {
         return hash.digest('hex');
     }
 
+    //cannot rely on nativeId as windows might leave a group after they are closed.
+    private getWindowGroupId = (identity: Identity): string => {
+        return `${identity.uuid}:${identity.name}`;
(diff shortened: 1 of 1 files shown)


src/browser/window_groups.ts
```diff
@@ -72,6 +72,11 @@ export class WindowGroups extends EventEmitter {
         return hash.digest('hex');
     }
 
+    //cannot rely on nativeId as windows might leave a group after they are closed.
+    private getWindowGroupId = (identity: Identity): string => {
+        return `${identity.uuid}:${identity.name}`;
```
datamadic: have these already been encoded so that `:` cant cause a problem?

nicholasdgoodman: also, dont we in other places use `/` when making UUID/name as a single string? Ought this not be standardized (including escaping or replacing the separator char)?

rdepena:
@pbaize updated based on review.

PR #682, opened 2018-12-14
RUN-4786 fixed remote subscriptions for System

whyn07m3:
ℹ️ This PR fixes timestamps for System's remote event subscriptions

➕ New automated test: [Multi-runtime missed events (System)](https://testing-dashboard.openfin.co/#/app/tests/5c129c05cb360141a7dfd6c4/edit)

### DIFF
--- a/src/browser/of_events.ts
+++ b/src/browser/of_events.ts
@@ -1,11 +1,13 @@
 import { app } from 'electron';
 import { EventEmitter } from 'events';
+import { isFloat } from '../common/main';
 import route from '../common/route';
 
 interface PastEvent {
     payload: any;
     routeString: string;
-    timestamp: number;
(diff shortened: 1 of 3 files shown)


tgoc99:
Do we need both times?  If we think one may be off by up to 15s then we could end up doubling down on some events.

whyn07m3:
@tgoc99 We need both because
1. _js-adapter_ adds timestamps, and since node won't have access to the native time function, it's using `Date.now()`. And since `Date.now()` has different signature and will always be bigger than native one, we will never emit missed events for node adapter
2. Time discrepancies happen sometimes on maybe even on some machines
3. I noticed big time differences only between JS timestamp and native timestamp in electron's logging function. The time difference before between different runtimes was usually pretty small

PR #685, opened 2018-12-17
RUN-4855 fixed multi-runtime propagation

pbaize:
Don't re-propagate events from an external runtime to avoid double propagation


pbaize:
@datamadic [this](https://github.com/HadoukenIO/core/pull/685/commits/178be2e089508515b7fb380e12e923adf9f4c40b) should solve the ordering, waiting on tests

PR #691, opened 2018-12-20
RUN-4596, RUN-4742: Flag-protected cache migration

anguslim:
This won't do anything unless old-local-storage-path, new-local-storage-path,
and local-storage-url are specified, in which case either the RVM
determined cache migration should be done by the runtime, or the user very
deliberately wanted the runtime to do cache migration.

See openfin/runtime#1250 for gory details.


index.js
```diff
@@ -514,6 +516,23 @@ function rvmCleanup(argo) {
     }
 }
 
+function migrateLocalStorage(argo) {
+    const oldLocalStoragePath = argo['old-local-storage-path'] || false;
+    const newLocalStoragePath = argo['new-local-storage-path'] || false;
+    const localStorageUrl = argo['local-storage-url'] || false;
```
HarsimranSingh: The types here are confusing:
```
const foo = string || boolean;
```

PR #695, opened 2018-12-27
RVM-674: Add blank security realm to runtime => RVM payloads

anguslim:
i.e. if no security realm is specified, add the field as an empty string.

turns out JSON.stringify() ignores all keys with undefined values


src/browser/rvm/rvm_message_bus.ts
```diff
@@ -278,7 +278,7 @@ export class RVMMessageBus extends EventEmitter  {
         const payload: any = Object.assign({
             processId: process.pid,
             runtimeVersion: processVersions.openfin,
-            securityRealm: argo ? argo['security-realm'] : ''
+            securityRealm: argo ? (argo['security-realm'] || '') : ''
```
HarsimranSingh: No need to check if `argo` is defined; it's one of the first things defined in core_state. `argo['security-realm'] || ''` should suffice.

anguslim: It's not defined in tests, so I added the extra check it in the original PR, #653

PR #699, opened 2019-01-03
RUN-4893 fix(v11 core): upgrade rx to rxjs

anguslim:
just what the doctor ordered (get it? because `rx`? ha)

implicit here is that an internet connection exists when building the core (because rxjs doesn't distribute a .min.js file, instead relying on [unpkg](https://unpkg.com) to do it)

this was followed, which was mostly painless: https://github.com/ReactiveX/rxjs/blob/4ffbbe56ab6c5ba46497c0154d83d27f882476fb/MIGRATION.md

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,6 +9,7 @@
 const fs = require('fs');
 const path = require('path');
 const asar = require('asar');
+const https = require('https');
 const electronRebuild = require('electron-rebuild');
 const wrench = require('wrench');
 const openfinSign = require('openfin-sign'); // OpenFin signing module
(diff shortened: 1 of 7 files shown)


nicholasdgoodman:
Does it have to perform the `GET` call on build? There are a number of scenarios I can think of where our own people may not have internet access while building the core (on a train, at a customer site, etc). Perhaps this logic would better exist in a `postinstall` action from the core's _package.json_.

datamadic:
does [this](https://github.com/reactivex/rxjs#es6-via-npm) not make it possible to just install it as dep like before?

PR #702, opened 2019-01-14
RUN-4933 Transition socket_server.js transport to socket_server.ts

nicholasdgoodman:
#### Description of Change

Convert `socket_server.js` to Typescript. Changes were done in stages:
- covert original JS prototypical class to ES6 JS class
- update import syntax, added types for websocket library (ws)
- full conversion to TypeScript

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Transition `socket_server.js` to TypeScript. No functional changes intended.


nicholasdgoodman:
@pbaize the vast majority of these functions are inherited from `ApiTransportBase<T>` and as-such cannot be declared as instance properties without also modifying the base class and the other derived implementation `ElipcStrategy`.

If you want I can convert to bound functions, but would have to do it in all three files.

Also, any idea why `ApiTransportBase` is generic to begin with? The only two derived classes have the same generic type argument.

PR #704, opened 2019-01-14
Run 4037

pbaize:
#### Description of Change
This PR adds enforcement of uuid uniqueness cross runtime without a dependence on the mesh.
- Runtimes in a security realm without the --enable-mesh flag are excluded from this behavior
- This will prevent the start up and connection of application/external applications if the uuid is taken
- Uses the named-mutex module to reserve an id on run
- This behavior is only guaranteed if all runtimes are >= the version this lands in.
- raises run-requested on the running app
- cleans non-running duplicated uuid from core state
 
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes:


whyn07m3:
⚠️ test runner stuck in a "failure" loop because most tests keep failing (not sure why). adding the label to ignore testing...

src/browser/api/application.js
```diff
@@ -747,17 +749,24 @@ function run(identity, mainWindowOpts, userAppConfigArgs) {
 }
 
 /**
- * Run an application via RVM
+ * Run an application via RVM. Returns a promise which never resolves, only rejects.
```
datamadic: Lets hold off on this one till your back. Though its little mem-wise, we should find a way to have this not hang/leak on success and be a bit more intuitive (success is the clean path, not reject)

src/browser/uuid_availability.ts
```diff
@@ -0,0 +1,22 @@
...
+        !meshEnabled || (
+            namedMutex.tryLock(makeMutexKey(uuid)) === 0
+        )
+    );
+}
+
+export function releaseUuid (uuid: string) {
+    const key = makeMutexKey(uuid);
+    let released = namedMutex.releaseLock(key);
+    while (!released) {
```
datamadic: Apologies for not looking into the namedMutex api ahead of time, but can you say a few nice words regarding `namedMutex.releaseLock` and how we are guaranteed to have it be released before too long? Just want to make sure that `while (!released)` can't get into a forever loop.

index.js
```diff
@@ -590,13 +597,23 @@ function launchApp(argo, startExternalAdapterServer) {
...
+                deleteApp(uuid);
+                duplicateUuidTransport.broadcast({ argv, uuid });
+                failedMutexCheck = true;
+            } else {
+                passedMutexCheck = true;
+            }
+        }
         // comparing ofManifestUrl and configUrl shouldn't consider query strings. Otherwise, it will break deep linking
-        if (startupAppOptions && (!isRunning || ofManifestUrl.split('?')[0] !== configUrl.split('?')[0])) {
+        if (startupAppOptions && passedMutexCheck && (!isRunning || ofManifestUrl.split('?')[0] !== configUrl.split('?')[0])) {
```
datamadic: ⚓️ :neckbeard:  this conditional, woof :)

index.js
```diff
@@ -590,13 +597,23 @@ function launchApp(argo, startExternalAdapterServer) {
 
         // this ensures that external connections that start the runtime can do so without a main window
         let successfulInitialLaunch = true;
-
+        let passedMutexCheck = false;
+        let failedMutexCheck = false;
+        if (uuid && !isRunning) {
+            if (!isUuidAvailable(uuid)) {
+                deleteApp(uuid);
```
datamadic: Which case is this, the uuid is in the local core state but not running but exists somewhere else? Aren't we cleaning them out when they close now?

src/browser/api_protocol/api_handlers/application.js
```diff
@@ -361,10 +365,29 @@ function runApplication(identity, message, ack, nack) {
     if (manifestUrl) {
         addRemoteSubscription(remoteSubscription).then((unSubscribe) => {
             remoteSubscriptionUnSubscribe = unSubscribe;
-            Application.runWithRVM(identity, manifestUrl).catch(nack);
+            unsub = duplicateUuidTransport.subscribeToUuid(uuid, (e) => {
```
datamadic: Is this assignment a race against lines 357 above in any way? They are both async no?

src/browser/api_protocol/api_handlers/authorization.js
```diff
@@ -159,7 +160,7 @@ function addPendingAuthentication(uuid, token, file, sponsor, authReqPayload) {
 }
 
 function authenticateUuid(authObj, authRequest, cb) {
-    if (ExternalApplication.getExternalConnectionByUuid(authRequest.uuid) || coreState.getAppByUuid(authRequest.uuid)) {
+    if (ExternalApplication.getExternalConnectionByUuid(authRequest.uuid) || coreState.getAppByUuid(authRequest.uuid) || !(authRequest.runtimeClient || isUuidAvailable(authRequest.uuid))) {
```
datamadic: This is the way to go, all of our external adapts have to send `runtimeClient` right?

src/browser/transport.ts
```diff
@@ -2,8 +2,36 @@ import Base from './transports/base';
...
 export { Base, ChromiumIPC, UnixDomainSocket, WMCopyData };
+
+export class NamedOneToManyTransport extends EventEmitter {
+    protected _transport: Base;
+
+    constructor(private name: string) {
+        super();
+    }
+
+    protected construct(): Base {
```
datamadic: fat arrow to protect `this`?

src/browser/duplicate_uuid_delegation.ts
```diff
@@ -0,0 +1,67 @@
...
+        } catch (e) {
+            log.writeToLog('info', `Duplicate Uuid delegation failed: ${JSON.stringify(e)}`);
+        }
+    }
+}
+
+async function subscribeToRunningExternal() {
+    //clean up apps after they are started in other runtimes
+    await subscribeToAllRuntimes({ listenType: 'on', className: 'system', eventName: 'application-started' });
+    ofEvents.on(route.system('application-started'), (data) => {
```
datamadic: This lives for the life of the browser process right?

src/browser/api_protocol/api_handlers/authorization.js
nicholasdgoodman: No other adapters send this property, so it will be `undefined`.

(This is a funny flag, considering there was a pre-existing property on the auth request: `authRequest.client.type` which has values like `'dotnet'` or `'java'`. It might not hurt to start setting `authRequest.client.type = 'runtime'` for consistency. Too late to change it now.)

src/browser/uuid_availability.ts
pbaize: @anguslim

src/browser/duplicate_uuid_delegation.ts
pbaize: yes

index.js
pbaize: This is in case we didn't clear it out. We only provide a new api call to clear out, we don't do it on close yet.

src/browser/api_protocol/api_handlers/application.js
pbaize: This will always happen before line 357, but I have the typeof check above to be sure

src/browser/transport.ts
pbaize: I just copied this over from PortDiscovery. can change to fat arrow if we want.

index.js
datamadic: gotcha 👍

src/browser/api_protocol/api_handlers/application.js
datamadic: 👍

src/browser/transport.ts
datamadic: ⚓️ :neckbeard:

src/browser/uuid_availability.ts
datamadic: So in the case where we successfully release the lock (no errors encountered) we will try to release it one more time and let the error code or -1 break the loop? Apologies if I'm missing something.
```
    let released = namedMutex.releaseLock(key); // = 0 == false
    while (!false) {
        released = namedMutex.releaseLock(key); // = some error code that breaks the loop?
    }


```

anguslim: Assuming no errors are encountered, the while loop will run until a release fails. As soon as the lock is completely released (i.e. the lock was released X times after being acquired X times), the next release attempt will result in an error, either within the Windows API (so some Windows error code is returned) or because the mutex has been cleaned up and no longer exists (so -1 is returned).

datamadic: ahh gotcha, thanks!

nicholasdgoodman: this seems like a rather un-standard way to use a mutex, maybe even has a bit of a code smell to it. is this the only place in code that the `releaseLock` is called? if not, doing so can cause breaks elsewhere.

the standard pattern is to call release once per acquire. im also curious as to why we would be attempting to acquire a mutex that is already held (i.e. multiple times). this is typically only done when passing through a synchronous callstack and each acquire ends with a release.

pbaize: The way the code is written the mutex will only be acquired once. The while is there as a backup. We did encounter a case where even though the release call returned ok, the mutex hadn't been deleted hence keeping the while.

src/browser/uuid_availability.ts
```diff
@@ -0,0 +1,22 @@
...
+const namedMutex = electron.namedMutex;
+import { meshEnabled } from './connection_manager';
+import { getAppRunningState } from './core_state';
+
+const makeMutexKey = (uuid: string) => `uuid-${uuid}`;
+
+export function isUuidAvailable(uuid: string) {
+    return !getAppRunningState(uuid) && (
+        !meshEnabled || (
+            namedMutex.tryLock(makeMutexKey(uuid)) === 0
```
nicholasdgoodman: i just looked at the implementation of this `tryLock` function in the Runtime.
it appears this is simply returning the raw value of the OS wait call. (@anguslim confirm?)

be-warned, value of `WAIT_OBJECT_0` (0) *and* `WAIT_ABANDONED` (128) indicates that the caller now owns the mutex.

src/browser/uuid_availability.ts
```diff
@@ -0,0 +1,22 @@
+const electron = require('electron');
+const namedMutex = electron.namedMutex;
+import { meshEnabled } from './connection_manager';
+import { getAppRunningState } from './core_state';
+
+const makeMutexKey = (uuid: string) => `uuid-${uuid}`;
+
+export function isUuidAvailable(uuid: string) {
```
nicholasdgoodman: just stylistically, i would use a stronger verb in the function name here. `isSomething` appears to be just passive "state check" call, but internally this function invokes `tryLock` which affects state globally across the entire system. a stronger name might be `tryReserveUuid` or something.

whyn07m3:
@pbaize Please, check if it runs through a suite (testing bot can't test it)

PR #706, opened 2019-01-15
RUN-4939 feat: move OS X single-tenant implementation to native code

anguslim:
renderers and GPU processes start before the core code that sets the cache path
so they don't log to debug.log as expected

requires native PR in https://github.com/openfin/runtime/pull/1331


anguslim:
Rebased

PR #707, opened 2019-01-16
RUN-4644 Replace Application "create" + "run" calls with "start" and "close" with "quit"

pbaize:
#### Description of Change
This adds a call to 'destroy-application' to the core to remove an app from core-state if it is not running. This is a prerequisite to [this PR](https://github.com/HadoukenIO/js-adapter/pull/230).

#### Checklist

- [x] `npm test` passes
- [x] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter)

### DIFF
--- a/src/browser/api/application.js
+++ b/src/browser/api/application.js
@@ -278,14 +278,14 @@ Application.close = function(identity, force, callback) {
         Window.close(mainWindowIdentity, force, callback);
     }
 };
-Application.destroy = function (identity, cb, nack) {
+Application.destroy = function(identity, cb, nack) {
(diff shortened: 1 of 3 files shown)


src/browser/api/application.js
```diff
@@ -278,7 +278,14 @@ Application.close = function(identity, force, callback) {
         Window.close(mainWindowIdentity, force, callback);
     }
 };
-
+Application.destroy = function(identity, cb, nack) {
```
datamadic: if this is an ack maybe lets rename `cb` to `ack`. Also might want to just `return` or `throw` here if this is backing an API call to be consistent with the new pattern.

PR #709, opened 2019-01-17
Check all elements in path, run on DOMContentLoaded

pbaize:
#### Description of Change
Updates the disabling of unsupported click types to prevent rogue child windows. It is still possible to click into a cross-origin iframe and have that Iframe open a child window which won't have the OpenFin api, but that is of less immediate concern.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Fixed an issue where clicking a link while holding a modifier click would open a child window which in tern could cause undesirable behavior.

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -3,6 +3,7 @@
 
 // THIS FILE GETS EVALED IN THE RENDERER PROCESS
 (function() {
+    console.log('in api_decorator');
     const glbl = global;
     const QUEUE_COUNTER_NAME = 'queueCounter';
     const noteGuidRegex = /^A21B62E0-16B1-4B10-8BE3-BBB6B489D862/;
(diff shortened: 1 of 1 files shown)


src/renderer/api-decorator.js
```diff
@@ -196,17 +197,21 @@
...
+                const tag = target.tagName;
+                const mightOpenNewWindow = tag === 'A' || tag === 'IMG';
+                if (mightOpenNewWindow && modifiedClick) {
                     e.preventDefault();
-                }
+                    return true;
+                } else if (modifiedClick && (tag === 'BUTTON' || tag === 'INPUT')) {
+                    if (target.type === 'submit') {
+                        e.preventDefault();
+                        return true;
```
datamadic: Could this prevent a form from being submitted?

src/renderer/api-decorator.js
```diff
@@ -3,6 +3,7 @@
 
 // THIS FILE GETS EVALED IN THE RENDERER PROCESS
 (function() {
+    console.log('in api_decorator');
```
datamadic: :neckbeard:  remove log

src/renderer/api-decorator.js
```diff
@@ -245,18 +250,18 @@
...
+    });
+    global.addEventListener('load', function() {        //---------------------------------------------------------------
         // TODO: extract this, used to be bound to ready
         //---------------------------------------------------------------
 
         // The api-ready event allows the webContents to assign api priority. This must happen after
         // any spin up windowing action or you risk stealing api priority from an already connected frame
         electron.remote.getCurrentWebContents(renderFrameId).emit('openfin-api-ready', renderFrameId);
-
+
```
datamadic: :neckbeard:  whitespace

src/renderer/api-decorator.js
pbaize: Possibly, we already do it though...

datamadic: indeed. ⚓️ 🐛. Are the added `return`s intentional?

pbaize: way to abort the iteration over path

PR #710, opened 2019-01-17
RUN-4813 RUN-4814 RUN-4815 Propagate resizes through the window graph

datamadic:
#### Description of Change

When a window resizes and is part of a group, the resize will now be
propagated through the group impacting windows that are "downstream"
of the resizing window in both the x and y directions. Resize events
along one axis or the other should now only impact windows that are
reachable through the window graph on that axis.

There are still cases where with this change I am seeing broken edges
on occation, in talks with the native team it may be due to scaling
(using dip or not). This exists currently and can be addressed in the
future if need be.

#### Checklist

- [X] `npm test` passes
- [X ] PR title starts with the JIRA ticket [pull request process](https://github.com/openfin/Internal-Wiki/wiki/Pull-Request-Process)
- [X ] PR release notes describe the change in a way relevant to app-developers

#### Release Notes
Window resizes in a group now propagate through the group. 

Notes:

### DIFF
--- a/src/browser/disabled_frame_group_tracker.ts
+++ b/src/browser/disabled_frame_group_tracker.ts
@@ -159,6 +159,7 @@ function handleBoundsChanging(
     changeType: ChangeType
 ): Move[] {
     const initialPositions: Move[] = getInitialPositions(win);
+    const leaderRectIndex = initialPositions.map(x => x.ofWin).indexOf(win);
     let moves: Move[];
     const startMove = moveFromOpenFinWindow(win); //Corrected
     const start = startMove.rect;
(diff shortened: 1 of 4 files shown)


datamadic:
yes(ish) on the src map. I can remove if we have concerns. Being able to step through the .ts files on unit tests is pretty nice though (as are stack traces to the real code). But if it adds too much space (adds 1.1M total) on the deploy or we just don't want it ill remove.

src/browser/disabled_frame_group_tracker.ts
```diff
@@ -323,7 +276,7 @@ export function addWindowToGroup(win: OpenFinWindow) {
                     } catch (error) {
                         writeToLog('error', error);
                     }
-                }, 16);
+                }, 30);
```
pbaize: Is there a particular reason for the timing change?

datamadic: Yes, just in testing it seemed that ~30 "fps" on the update was more than adequate for smooth resize so figured why not be a little kinder on the core.

PR #715, opened 2019-02-05
NWI initial intgration

datamadic:
#### Description of Change
This is the first step to adding native window support. This represents the abstraction of actions taken on a browser window to allow for external windows to be layered in later. On its own it will not add any functionality. 

@whyn07m3 please have a look and if you could pick up on any PR feedback it would be much appreciated. 

[tests well](https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5c59e786cb360141a7dfdb0b)

#### Checklist

- [x] `npm test` passes

#### Release Notes
n/a

### DIFF
--- a/src/browser/api/native_window.ts
+++ b/src/browser/api/native_window.ts
@@ -1,0 +1,217 @@
+import { Bounds } from '../../../js-adapter/src/shapes';
+import { BrowserWindow, Rectangle } from 'electron';
+import { clipBounds, windowSetBoundsToVisible } from '../utils';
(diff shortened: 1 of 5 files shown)


rdepena:
Not really a blocker but this could use some [mockery style](https://github.com/HadoukenIO/core/blob/develop/test/file_download.ts) Unit tests

rdepena:
Not sure if I agree with the approach here, can we have a live session @whyn07m3 @datamadic, I might just fundamentally not understand the change set.

PR #722, opened 2019-02-14
RUN-4958 Warning in console if no license key is included

licui3936:
#### Description of Change
This PR is to check the license key field in the app manifest and also does validation that the length of license key should be 36 characters. If the license key is missing or the length is wrong, we will display a warning message in the console. 

#### Checklist
- [x] `npm test` passes

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -33,7 +33,8 @@
         options: { api: { iframe: { enableDeprecatedSharedName } } },
         socketServerState,
         runtimeArguments,
-        frames
+        frames,
+        licenseKey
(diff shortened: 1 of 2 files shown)


src/renderer/api-decorator.js
```diff
@@ -310,7 +318,10 @@
 
 
     function onContentReady(bindObject, callback) {
-
+        if (!isLicenseKeyValid()) {
+            console.warn('WARNING : Application does not have a valid OpenFin license key implemented in application manifest.' +
```
pbaize: Missing a space between these sentences. Maybe use a template string?

PR #730, opened 2019-02-21
RUN-5006 fixes core exception for resizeRegion

whyn07m3:
#### Description of Change
This fixes core exception thrown in some cases when updating window's `resizeRegion` option before the window is shown but immediately after it is created.

#### Checklist
- [x] `npm test` passes
- [x] automated test added: [Core exception: Window.updateOptions (resizeRegion)](https://testing-dashboard.openfin.co/#/app/tests/5c6db59ecb360141a7dfdbd0/edit)

#### Release Notes
Bug fix: exception thrown in some cases when updating window's `resizeRegion` option.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1956,7 +1956,7 @@ function handleForceActions(identity, force, eventType, eventPayload, defaultAct
 
 
 function applyAdditionalOptionsToWindow(browserWindow) {
-    let options = browserWindow && browserWindow._options;
+    const options = browserWindow && JSON.parse(JSON.stringify(browserWindow._options));
 
     if (!options) {
         return;


src/browser/api/window.js
```diff
@@ -1956,7 +1956,7 @@ function handleForceActions(identity, force, eventType, eventPayload, defaultAct
 
 
 function applyAdditionalOptionsToWindow(browserWindow) {
-    let options = browserWindow && browserWindow._options;
+    const options = browserWindow && JSON.parse(JSON.stringify(browserWindow._options));
```
pbaize:
```suggestion
    const options = browserWindow && Object.assign({}, browserWindow._options);
```

pbaize: `JSON.parse(JSON.stringify(undefined))` raises an exception.

whyn07m3: I agree with you about a possible exception thrown, but it's not expected, because we set `._options` just 2 lines above: https://github.com/HadoukenIO/core/blob/3147096ffa3c353d442058a152d50f24f75d0924/src/browser/api/window.js#L497

I can't use `Object.assign` as I need a deep clone.

datamadic: Can we just double check using getters / setters to log out access as this is async essentially and maybe, just maybe, something is getting in between and tripping us up here.

PR #736, opened 2019-02-27
log error message if dup uuid transport listener not unhooked properly

datamadic:
log error message if dup uuid transport listener not unhooked properly

### DIFF
--- a/src/browser/api_protocol/api_handlers/application.js
+++ b/src/browser/api_protocol/api_handlers/application.js
@@ -336,7 +336,10 @@ function runApplication(identity, message, ack, nack) {
     const appIdentity = apiProtocolBase.getTargetApplicationIdentity(payload);
     const { uuid } = appIdentity;
     let remoteSubscriptionUnSubscribe;
-    let unsub = () => null;
+    const logMsg = manifestUrl ?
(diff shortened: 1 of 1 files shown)


src/browser/api_protocol/api_handlers/application.js
```diff
@@ -336,7 +337,10 @@ function runApplication(identity, message, ack, nack) {
     const appIdentity = apiProtocolBase.getTargetApplicationIdentity(payload);
     const { uuid } = appIdentity;
     let remoteSubscriptionUnSubscribe;
-    let unsub = () => null;
+    const logMsg = manifestUrl ?
+        `unsub called before duplicate uuid transport listener removed for ${uuid}` :
+        '';
+    let unsub = () => writeToLog('error', logMsg);
```
HarsimranSingh: :neckbeard: `const`

datamadic: that will get reassigned further down

HarsimranSingh: Yup i'm blind https://github.com/HadoukenIO/core/pull/736/files#diff-dd883de5703030a371af2cdd9ad72616R376

PR #741, opened 2019-02-28
RUN-5006 deep merge of window options

whyn07m3:
### Description of Change
Window options updates will now be extending previous options, instead of replacing them.
This was causing an exception in the core for this specific scenario:
1. Create a new window with `autoShow: false` and `frame: false`
2. Partially update window options (like in the example below)
3. Show the window
4. Error box shown for the exception thrown in the core

#### Old behavior example:
1. Window options before updating: `{ ..., resizeRegion: { ..., sides: { top: true, ... }}}`
2. `wnd.updateOptions({ resizeRegion: { sides: { top: false } } })`
3. Window options after updating: `{ resizeRegion: { sides: { top: false } } }` 

#### New behavior example (this PR):
1. Window options before updating: `{ ..., resizeRegion: { ..., sides: { top: true, ... }}}`
2. `wnd.updateOptions({ resizeRegion: { sides: { top: false } } })`
3. Window options after updating: `{ ..., resizeRegion: { ..., sides: { top: false, ... }}}` 

#### Docs changes:
1. [js-adapter](https://github.com/HadoukenIO/js-adapter/pull/250)
2. [javascript-adapter](https://github.com/openfin/javascript-adapter/pull/527)

#### Checklist
- [x] `npm test` passes
- [x] automated test added: [Core exception: Window.updateOptions (resizeRegion)](https://testing-dashboard.openfin.co/#/app/tests/5c6db59ecb360141a7dfdbd0/edit)

#### Release Notes
Bug fix: exception thrown in some cases when partially updating window's options.

### DIFF
--- a/src/common/main.ts
+++ b/src/common/main.ts
@@ -107,3 +107,35 @@ export function noop(): void {
 export function isFloat(n: any): boolean {
     return Number(n) === n && n % 1 !== 0;
 }
+
+export function isObject(item: any): boolean {
+    return (item && typeof item === 'object' && !Array.isArray(item));
+}
+
+// Deep merge https://stackoverflow.com/a/34749873
+export function mergeDeep(target: any, ...sources: any[]): any {
+    if (!sources.length) {
+        return target;
+    }
+
+    const source = sources.shift();
+
+    if (isObject(target) && isObject(source)) {
+        const keys = Object.keys(source);
+
+        for (let i = 0; i < keys.length; i++) {
+            const key = keys[i];
+
+            if (isObject(source[key])) {
+                if (!target[key]) {
+                    Object.assign(target, { [key]: {} });
+                }
+                mergeDeep(target[key], source[key]);
+            } else {
+                Object.assign(target, { [key]: source[key] });
+            }
+        }
+    }
+
+    return mergeDeep(target, ...sources);
+}
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -35,7 +35,7 @@ import { toSafeInt } from '../../common/safe_int';
 import route from '../../common/route';
 import { FrameInfo } from './frame';
 import { System } from './system';
-import { isFileUrl, isHttpUrl, getIdentityFromObject } from '../../common/main';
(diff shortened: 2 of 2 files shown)


src/common/main.ts
```diff
@@ -107,3 +107,35 @@ export function noop(): void {
 export function isFloat(n: any): boolean {
     return Number(n) === n && n % 1 !== 0;
 }
+
+export function isObject(item: any): boolean {
+    return (item && typeof item === 'object' && !Array.isArray(item));
+}
+
+// Deep merge https://stackoverflow.com/a/34749873
+export function mergeDeep(target: any, ...sources: any[]): any {
```
datamadic: we also do this [here](https://github.com/HadoukenIO/core/blob/develop/src/browser/convert_options.js#L151). If we are doing the same thing in 2 places lets use the same functions to do it.

datamadic:
note: the failed tests don't apply here. Any re-tests (if needed) should target canary.

src/common/main.ts
whyn07m3: The function you mentioned (`validate`) does these things differently:
1. **Important difference**: it only sets overwrites base values if both types of base & new values match. If they don't, it'll keep the base value.
2. **Minor difference**: It also only checks if the value is an object, but it could be an array. I know we don't have arrays in windows options, but...

datamadic: For 1, isn't that what we want? do we have cases where the thing you set it to is a different type than the default?
For 2, I'd argue that merging an array is not a thing we ever want to do

whyn07m3: For 1, I'm afraid switching that method to use newly added `deepMerge`, because deepMerge will simply overwrite provided values. In this function, we only do that if their types match. It sounds risky to me because they slightly do different things.

datamadic: why not use that method in place of deep merge?

whyn07m3: I'm not sure about all possible scenarios of updating window options. For example, we used to have preloadScripts (and plugins, I think) with different value types, and it wouldn't work in those cases where there are different value types. So if we re-use current method, we would have to let users know that the options are not being "extended, only for matching types.

PR #743, opened 2019-03-06
Check if name passed to window.open()

aziz512:
#### Description of Change
Fixes a bug where not passing a window name to `window.open()` (like `window.open('https://example.com')` ) ignores custom options (i.e. `waitForPageLoad: false`).

#### Checklist
- [x] `npm test` passes

#### Release Notes

Allows to use `window.open()` without passing a window name.

### DIFF
--- a/src/renderer/api-decorator.js
+++ b/src/renderer/api-decorator.js
@@ -340,7 +340,7 @@
         const [url, requestedName, features = ''] = args; // jshint ignore:line
         const requestId = ++childWindowRequestId;
         const webContentsId = getWebContentsId();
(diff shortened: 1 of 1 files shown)


licui3936:
need to fix the broken test.

datamadic:
This is likely going to be folded into a larger fix that Li is working on (taking this code as part). I think we owe you a lunch for all the code we are taking but not merging directly 😄.

datamadic:
Closing this for now. The changes have been absorbed into Li's fix which will address the failed test. 

Again, apologies for all the closings, this is good work!

PR #747, opened 2019-03-14
Child windows not created with the OpenFin API will have waitForPageLoad set to false

luiemilio:
#### Description of Change
Child window creation now checks if it the call did not come from our API. If it doesn't, then it will set waitForPageLoad to false for these calls.

#### Checklist

- [X] `npm test` passes

#### Release Notes

Notes:


datamadic:
This is going to get taken care of with Li's PR. Going to close for now

PR #754, opened 2019-03-27
RUN-5069 Add window options updated event (take #2)

tomer-openfin:
#### Description of Change
**Added a 'options-changed' event to window after window.updateOptions is done.** 

The payload contains a diff array containing prev and new values for all changed options.

Would appreciate if you could also take a look at the [dashboard test](https://testing-dashboard.openfin.co/#/app/tests/5c9a5c6acb360141a7dfdec4/edit)

Related [js adapter PR](https://github.com/HadoukenIO/js-adapter/pull/259)
#### Checklist

- [x] PR description included
- [x] `npm test` passes
- [x] relevant documentation was added

#### Release Notes

Notes: Added a 'options-changed' event to window after window.updateOptions is done.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1697,12 +1697,31 @@ Window.defineDraggableArea = function() {};
 
 Window.updateOptions = function(identity, updateObj) {
     let browserWindow = getElectronBrowserWindow(identity, 'update settings for');
+    let { uuid, name } = identity;
+    let diff = {},
+        invalidOptions = [];
+    let clone = obj => JSON.parse(JSON.stringify(obj)); // this works here, but has limitations; reuse with caution.
 
     try {
         for (var opt in updateObj) {
+
             if (optionSetters[opt]) {
+                let oldVal = clone(getOptFromBrowserWin(opt, browserWindow));
                 optionSetters[opt](updateObj[opt], browserWindow);
+                let newVal = clone(getOptFromBrowserWin(opt, browserWindow));
+
+
+                if (!_.isEqual(oldVal, newVal)) {
+                    diff[opt] = { oldVal, newVal };
+                }
+            } else {
+                invalidOptions.push(opt);
             }
+        }
+
+        let options = browserWindow && clone(browserWindow._options);
+        if (Object.keys(diff).length) {
+            ofEvents.emit(route.window('options-changed', uuid, name), { uuid, name, options, diff, invalidOptions });
         }
     } catch (e) {
         console.log(e.message);


src/browser/api/window.js
```diff
@@ -1697,13 +1697,21 @@ Window.defineDraggableArea = function() {};
...
+    let { uuid, name } = identity;
+    let diff = [];
 
     try {
         for (var opt in updateObj) {
             if (optionSetters[opt]) {
+                let optionValues = { optionName: opt };
+                optionValues.oldVal = getOptFromBrowserWin(opt, browserWindow);
                 optionSetters[opt](updateObj[opt], browserWindow);
+                optionValues.newVal = getOptFromBrowserWin(opt, browserWindow);
```
tgoc99: I think some options might be objects (see cornerRounding - https://cdn.openfin.co/docs/javascript/alpha/Window.html).  Do we need to clone/copy the oldVal/newVal here?

tomer-openfin: Pushed a fix.

tomer-openfin:
I can do that. Do you think it would be used? In other words - how often would a user care about one option when another change?

tomer-openfin:
Update - I'll keep the payload as it is for now. Will add the whole options object if the need arises.

tomer-openfin:
@datamadic @pbaize I improved the types and changed the payload to contain the options object and a list of invalid options.
Note that I'm firing the event even if the diff array is empty - in case someone would be interested in that.

Thanks for the great feedback (!) keep it coming.

datamadic:
Looking good. A few more comments :) Can we remove the JS adapter from this PR? Looks like it sneaked in [here](https://github.com/HadoukenIO/core/pull/754/commits/402c6ff0845183a3480903ea09afc035251eaaf1). Also, can we make the `diff` a map of the old values to make the typing more straight forward and access a bit simpler (no need to iterate / search for options in a list, reduce redundant `newVal` ). Also also, lets check that something actually changed before sending out that event, something like `if (Object.keys(diff).length > 0) {...}`

tomer-openfin:
**About the JS adapter** - sure, sorry about the mess.
**About the diff** - Not sure I get what you mean by 

> a map of old values

Do you mean an object with options names for keys? like so:
`{
      [optionsName]: {prevVal: 'val1', newVal: 'val2'}
}` ? or something else?

**About checking to see if the list isn't empty** - sure. Do you think users wouldn't care about listening to updateOptions calls regardless of change? also, should the event dispatch if bad props were passed? (diff is empty but invalidOptions has items)?

Thanks again for the feedback @datamadic

PR #772, opened 2019-04-25
RUN-5166 Add a custom lint rule in the core to prevent using 'fs' in favor of 'original-fs'

aziz512:
#### Description of Change
Using `fs.readFileSync` in the core will not work correctly because we will add asar verification, which requires every `filename.ext` being read to need a `filename.ext.ofds` file associated with it.

This PR adds a custom lint rule that prevents people from importing `readFileSync` from `fs`, prompting them to import it from `original-fs` instead.
It also sets up a framework for future custom lint rules.

#### Checklist
- [x] `npm test` passes
- [x] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter)

#### Examples
```
import { readFileSync } from 'fs'; // fail
import { appendFile, readFileSync } from 'fs' // fail
import readFileSync from 'fs'; // fail
import 'fs'; // pass
import { appendFile } from 'fs' // pass
```

@rdepena @HarsimranSingh


aziz512:
> do we need to go through and opt out now when it is appropriate to use `fs` or should this build fine?

@datamadic
This should (and does) build fine since the existing codebase doesn't contain any cases handled by this rule.

PR #773, opened 2019-04-26
RUN-5112 Add the ability to get the desktop owner settings for services

datamadic:
#### Description of Change
Adds the ability to get the json blob that is associated with a service. As written you can get any configuration you'd like provided you know it's name. We can handle discovery and if these should be private to the service itself in later efforts.

A note about testing: currently it is not feasible to round-trip test a custom version of the runtime and rvm simultaneously. This change requires just that. It also requires setting custom a custom desktop owner settings file compounding the testing difficulties. For now, lets please just use our brains in review as best we can :)

@pjbroadbent  

#### Checklist

- [x] `npm test` passes

Linked to [this js adapter pr](https://github.com/HadoukenIO/js-adapter/pull/275)

Note: this requires [this](https://github.com/openfin/openfinrvm/commit/57cc300a3989feb6e5324d53b8f13059f68b5328) RVM change to be released

### DIFF
--- a/src/browser/api_protocol/api_handlers/system.ts
+++ b/src/browser/api_protocol/api_handlers/system.ts
@@ -78,6 +78,7 @@ export const SystemApiMap: APIHandlerMap = {
     'get-remote-config': { apiFunc: getRemoteConfig, apiPath: '.getRemoteConfig' },
     'get-runtime-info': getRuntimeInfo,
     'get-rvm-info': getRvmInfo,
+    'get-service-configuration': getServiceConfiguration,
     'get-preload-scripts': getPreloadScripts,
     'get-version': getVersion,
     'launch-external-process': { apiFunc: launchExternalProcess, apiPath: '.launchExternalProcess' },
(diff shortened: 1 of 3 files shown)


src/browser/api_protocol/api_handlers/system.ts
```diff
@@ -115,6 +116,30 @@ function didFail(e: any): boolean {
...
+    if (didFail(response)) {
+        throw response;
+    }
+
+    if (!Array.isArray(response)) {
+        throw new Error(`Settings in desktop owner settings are not configured correctly, please see
+         ${dosURL} for configuration information`);
+    }
+
+    const config = response.find(service => service.name === name);
```
rdepena: Any reason we are doing filtering at the `handler` level versus the `api` level?

datamadic: I like throwing safely from there, want to keep the system call as safe as possible just in case we use it internally

tgoc99: Is this just for OF services?  I wonder if we could make this a more generic "ask RVM" API so that services (& other apps?) can take new RVM features without runtime upgrades.  RVM is making a big push to have everyone on latest, so if we make a generic API (and maybe not document it?) then RVM can expose some new features without runtime changes.  Just a thought, maybe not for this PR.

datamadic: Yes, in principle I like this. And because its whats under the covers we can swap this on top of something like that in the future.

PR #782, opened 2019-05-07
RUN-5200: Don't close window till content loaded - 10.66.41/staging

dhamberlin:
#### Description of Change
Duplicate of https://github.com/HadoukenIO/core/pull/781, but targeting 10.66.41/staging.

Jira ticket: https://appoji.jira.com/browse/RUN-5200

When user closes a window before it's safe, hide the window, but wait for the DOM `load` event is fired to actually close the window.

SIGNIFICANT CAVEAT: This works for most sites I tested, but NOT for cnn.com. I suspect iframe shenanigans, or possibly something caused by their very slow js event handlers. Further investigation needed

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -440,20 +440,16 @@ Window.create = function(id, opts) {
         ofEvents.emit(route.window('openfin-diagnostic/unload', uuid, name, true), url);
     };
 
-    // Hack: closing a window before content is finish loading in 10.66.40.* causes the renderer to crash.
-    // Hide the window instead, and close from the core when it's safe.
-    let wasCloseRequested = false;
-    const preventClose = (event) => {
-        wasCloseRequested = true;
-        browserWindow.hide();
-    };
-    ofEvents.on(route.window('close-requested', uuid, name), preventClose);
-    ofEvents.once(route.window('initialized', uuid, name), () => {
-        ofEvents.removeListener(route.window('close-requested', uuid, name), preventClose);
-        if (wasCloseRequested) {
-            Window.close({ uuid, name });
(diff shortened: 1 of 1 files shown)


dhamberlin:
Reverted to the version that ignores the iframe issue. @StevenEBarbaro drew my attention to some other webContents events that might allow us to provide a more comprehensive fix. But, as discussed, this does somewhat increase code complexity, potentially opening us up to unsavory edge cases. I'll play with that a little, but I suspect that this change may be an optimal balance of improving the current situation while minimizing the risk of inadvertently causing new bugs.

src/browser/api/window.js
```diff
@@ -440,6 +440,22 @@ Window.create = function(id, opts) {
...
+    let wasCloseRequested = false;
+    const preventClose = (event) => {
+        wasCloseRequested = true;
+        browserWindow.hide();
+    };
+    ofEvents.on(route.window('close-requested', uuid, name), preventClose);
+    ofEvents.once(route.window('initialized', uuid, name), () => {
+        ofEvents.removeListener(route.window('close-requested', uuid, name), preventClose);
+        if (wasCloseRequested) {
+            Window.close({ uuid, name });
```
rdepena: This should force close here. under normal circumstances any close-requested event would not fire (assuming the user closes the window before the context is ready).

datamadic: This would need to ensure that downstream `will-close` events fire then

rdepena: Talked offline, let's ignore that one.

PR #784, opened 2019-05-09
RUN-5213 -  Grouping based on will-move and will-resize

tgoc99:
#### Description of Change
These two events happen prior to a move or resize and you can prevent default on the event.  This is very similar to disabling “user-movement”.  It provides direct or indirect solution to the following tickets - in parens I put whether we can accomplish this with the old grouping paradigm without native intervention:

RUN-5213  - investigate will-move & will-resize (N/A)
RUN-5163 - move during disabled-frame-bounds-changing (No)
RUN-5067 - restore maximized tabset on move (No)
RUN-5168 - calling setbounds on window while moving not working (No)
RUN-5086 - behavior of maximized/minimized windows in groups (No)
RUN-5087 - get resize events on size constrained windows (Yes)
RUN-5074 - correct the changeType (Kind of)
RUN-5221  - move to delta based calcs (Yes)
???  - moving via right-click context menu (can only workaround)

We need to communicate this with the layouts team.  In this PR, the events are exposed with the physical pixel information but including monitor scaling in the payload.  (i.e. if you were to getBounds at 150% monitor scaling you would get a 500 pixel width but will-move and will-resize payloads would have 750 pixel width with 1.5 monitorScaling).  We should test this out on layouts ASAP. 

#### Checklist

- [ x ] PR description included and stakeholders cc'd
- [ x ] `npm test` passes
- [ x ] PR title starts with the JIRA ticket [pull request process](https://github.com/openfin/Internal-Wiki/wiki/Pull-Request-Process)
- [ x ] PR release notes describe the change in a way relevant to app-developers
- [ x ] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter) (events added in [this adapter PR](https://github.com/HadoukenIO/js-adapter/pull/281))
- [ ] manual tests are [changed or added](https://github.com/openfin/test_apps) (Layouts testing?)

#### Release Notes

Notes: will-move & will-resize events added for use when monitor scaling is not set to 100%.


src/browser/disabled_frame_group_tracker.ts
```diff
@@ -152,45 +125,30 @@ const makeTranslate = (delta: RectangleBase) => ({ ofWin, rect, offset }: Move):
...
-                : initialPositions;
             const moved = (xShift || yShift);
-            //This flag is here because sometimes the runtime lies and says we moved on a resize
-            //This flag should always be set to true when relying on runtime events. It should be false on api moves.
-            //Setting it to false on runtime events can cause a growing window bug.
-            moves = moved && !treatBothChangedAsJustAResize
-                ? handleMoveOnly(start, start.shift(shift), moves)
-                : moves;
-
+            if (resized) {
```
pbaize: Do we no longer hit this on user moves? this used to cause growing windows in certain conditions

src/browser/disabled_frame_group_tracker.ts
```diff
@@ -228,111 +186,70 @@ function handleResizeOnly(leaderRectIndex: number, startMove: Move, end: Rectang
...
-
-export function removeWindowFromGroup(win: OpenFinWindow) {
-    if (!win.browserWindow.isDestroyed()) {
-        win.browserWindow.setUserMovementEnabled(true);
-        const winId = win.browserWindow.nativeId;
-        const listener = listenerCache.get(winId);
-        if (listener) {
-            win.browserWindow.removeListener('disabled-frame-bounds-changing', listener);
+    const moveListener = (e: any, bounds: RectangleBase) => genericListener(e, bounds, 0);
+    const resizeListener = (e: any, bounds: RectangleBase) => genericListener(e, bounds, 1);
```
pbaize: Should this be changetype 2 when the origin of the window changes? That's current behavior

pbaize:
Overall looks good, just that changetype concern.

src/browser/disabled_frame_group_tracker.ts
tgoc99: Isn't that move and resize? I don't think its possible for that to happen. Will-resize and will-move events are only for user generated moves.

src/browser/disabled_frame_group_tracker.ts
tgoc99: Right, the changetype is being set to either 0 or 1 (move or resize) so this can only be hit on API calls.

src/browser/disabled_frame_group_tracker.ts
pbaize: No, it's change in position and size, so top-left resize is changetype 2

pbaize: I'm not concerned about the wrong changetype for our internal logic, but we should make sure we send the right info out on the event listeners

tgoc99: Not seeing that behavior with bounds-changing.  Seems like a bug if that is true with other events.

pbaize: It's true with disabled-frame-bounds-changing

tgoc99: spoke off-line, disabled-frame being changetype 2 is a bug

tgoc99:
This PR breaks Mac grouping - it was easier to fix with the code added in #796 that builds on this PR  to implement native grouping so closing this one in favor of a single PR.

PR #786, opened 2019-05-09
Run 5188/part 2

pbaize:
#### Description of Change
Follow up to part 1. Routes webcontents related apis separately so they can be shared with browserview

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: n/a

### DIFF
--- a/src/browser/api/webcontents.ts
+++ b/src/browser/api/webcontents.ts
@@ -1,6 +1,5 @@
 import * as url from 'url';
 
-
 export function executeJavascript(webContents: Electron.WebContents, code: string, callback: (e: any, result: any) => void): void {
     webContents.executeJavaScript(code, true, (result) => {
         callback(undefined, result);
@@ -18,7 +17,6 @@ export function getInfo(webContents: Electron.WebContents) {
 
 export function getAbsolutePath(webContents: Electron.WebContents, path: string) {
     const windowURL = webContents.getURL();
-
     return  url.resolve(windowURL, path);
 }
 
@@ -41,8 +39,7 @@ export async function navigateBack (webContents: Electron.WebContents) {
 
 export async function navigateForward (webContents: Electron.WebContents) {
     if (!webContents.canGoForward()) {
-        const error = new Error('Cannot navigate forward');
-        throw error;
+        throw new Error('Cannot navigate forward');
     }
     const navigationEnd = createNavigationEndPromise(webContents);
     webContents.goForward();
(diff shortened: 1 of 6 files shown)


src/browser/api/webcontents.ts
```diff
@@ -0,0 +1,86 @@
+import * as url from 'url';
+
+export function executeJavascript(webContents: Electron.WebContents, code: string, callback: (e: any, result: any) => void): void {
```
tomer-openfin: Maybe we want to use the opportunity to move to a Promise flow here?

src/browser/api/webcontents.ts
```diff
@@ -0,0 +1,86 @@
...
+        url: webContents.getURL()
+    };
+}
+
+export function getAbsolutePath(webContents: Electron.WebContents, path: string) {
+    const windowURL = webContents.getURL();
+    return  url.resolve(windowURL, path);
+}
+
+export function navigate (webContents: Electron.WebContents, url: string) {
```
MichaelMCoates: minor detail, but spacing consistency after function names. Feel free to ignore.

src/browser/api/window.js
```diff
@@ -35,6 +34,7 @@ import { toSafeInt } from '../../common/safe_int';
 import route from '../../common/route';
 import { FrameInfo } from './frame';
 import { System } from './system';
+import * as WebContents from './webcontents';
```
tomer-openfin: an opinionated note -imo it's nicer when there's 1 default export containing an object with all the methods in it, along with individual exports for the methods. That way this line turns to 

> import WebContents from './webcontents';

src/browser/api/window.js
```diff
@@ -1272,23 +1270,16 @@ Window.getGroup = function(identity) {
...
     return windowInfo;
 };
 
 
 Window.getAbsolutePath = function(identity, path) {
     let browserWindow = getElectronBrowserWindow(identity, 'get URL for');
-    let windowURL = browserWindow.webContents.getURL();
-
-    return (path || path === 0) ? url.resolve(windowURL, path) : '';
+    return (path || path === 0) ? WebContents.getAbsolutePath(browserWindow.webContents, path) : '';
```
tomer-openfin: why would path equal 0?

src/browser/api_protocol/api_handlers/webcontents.ts
```diff
@@ -0,0 +1,130 @@
...
+            });
+        }
+        pUuid = Application.getParentApplication({
+            uuid: pUuid
+        });
+    }
+
+    return nack(new Error('Rejected, target window is not owned by requesting identity'));
+}
+function navigateWindow(identity: Identity, message: APIMessage, ack: Acker, nack: (error: Error) => void): void {
```
tomer-openfin: Does the name still make sense after moving out of window?

tomer-openfin: (this applies to other methods here too)

pbaize: I'd rather rename in a follow up pr, for now it still only does windows and I'm trying to keep changes to a minimum

rdepena: No rename is needed, the API action cannot change `navigate-window` and the only purpose of the API name is to be mapped.

pbaize: @rdepena I assume @tomer-openfin was referencing the function names, not the api mapping

rdepena:
👍

src/browser/api_protocol/api_handlers/webcontents.ts
rdepena: right, but the function is only used by the mapping so it should just keep the name.

PR #790, opened 2019-05-15
RUN-5288 Force lower case for username on getDeviceUserId

luiemilio:
#### Description of Change

If the machine's username changes case on reboot, getDeviceUserId will return a different value each time. We should change it so we always lower case this value to handle this situation.

#### Checklist

- [ X] PR description included and stakeholders cc'd
- [ X] `npm test` passes
- [ X] PR title starts with the JIRA ticket [pull request process](https://github.com/openfin/Internal-Wiki/wiki/Pull-Request-Process)

#### Release Notes

Notes: 

Fixed issue where getDeviceUserId returned a different value if the machine's username switched between upper and lower case.


nicholasdgoodman:
Perhaps the real problem is we are getting the username from the wrong place:
https://github.com/HadoukenIO/core/blob/develop/src/browser/api/system.js#L241

Maybe replace:
```
username = process.env.USERNAME;  // This can be overwritten / change to anything you want!
```

with:
```
username = os.userInfo().username;
```

datamadic:
Lets hold off on this as it's a breaking change for anyone currently using it. I have added a more complete ticket, [RUN-5289](https://appoji.jira.com/browse/RUN-5289), to the next sprint. Luis, please elevate priority if you think this is more critical.

PR #799, opened 2019-05-24
RUN-5265 cleaned up external window info object

whyn07m3:
### Description of Change
This PR changes **some** of external window info objects returned to the client.
The changes affect the following API/events (these were unnecessarily getting full info object / extra info).
 Other APIs / events return only needed info, similar in shape to regular OpenFin windows.
- `fin.System.getAllExternalWindows`
- `"external-window-closed"`
- `"external-window-created"`
- `"external-window-hidden"`
- `"external-window-shown"`

Info object **before**:
```
alwaysOnTop: false,
  bounds: {
    height: 265,
    width: 319,
    x: 150,
    y: 350,
  },
  className: 'Notepad',
  dpi: 96,
  dpiAwareness: -1,
  focused: false,
  id: '0x00040658',
  maximized: false,
  minimized: false,
  name: 'Notepad',
  process: {
    imageName: 'C:\Windows\System32\notepad.exe',
    injected: false,
    pid: 5924,
  },
  title: 'Untitled - Notepad',
  uuid: '0x00040658',
  visible: true,
```

Info object **now**:
```
{
  name: 'Notepad',
  process: {
    injected: false,
    pid: 5924,
  },
  title: 'Untitled - Notepad',
  uuid: '0x00040658',
  visible: true,
}
```

**Note:** `externalWindow.getInfo()` still returns full info, except `process.imageName`.

### Checklist

### DIFF
--- a/src/browser/api/external_window.ts
+++ b/src/browser/api/external_window.ts
@@ -1,7 +1,7 @@
 import { app as electronApp, ExternalWindow, WinEventHookEmitter, NativeWindowInfo } from 'electron';
 import { Bounds } from '../../../js-adapter/src/shapes';
 import { EventEmitter } from 'events';
-import { extendNativeWindowInfo } from '../utils';
+import { getNativeWindowInfo, getNativeWindowInfoLite } from '../utils';
 import { Identity } from '../../../js-adapter/src/identity';
 import { OF_EVENT_FROM_WINDOWS_MESSAGE } from '../../common/windows_messages';
 import * as NativeWindowModule from './native_window';
@@ -77,7 +77,7 @@ export function getExternalWindowGroup(identity: Identity): Shapes.GroupWindowId
 export function getExternalWindowInfo(identity: Identity): Shapes.NativeWindowInfo {
   const { uuid } = identity;
   const rawNativeWindowInfo = electronApp.getNativeWindowInfoForNativeId(uuid);
-  return extendNativeWindowInfo(rawNativeWindowInfo);
+  return getNativeWindowInfo(rawNativeWindowInfo);
 }
 
 export function getExternalWindowOptions(identity: Identity): any {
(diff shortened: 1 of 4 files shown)


whyn07m3:
@tomer-openfin please, review

src/browser/api/external_window.ts
```diff
@@ -541,7 +541,8 @@ async function subscribeToInjectionEvents(externalWindow: Shapes.ExternalWindow)
 /*
     Decides whether external window is valid (external window filtering)
 */
-export function isValidExternalWindow(nativeWindowInfo: Shapes.NativeWindowInfo, ignoreVisibility?: boolean) {
+export function isValidExternalWindow(rawNativeWindowInfo: Shapes.RawNativeWindowInfo, ignoreVisibility?: boolean) {
+  const nativeWindowInfo = parseRawNativeWindowInfo(rawNativeWindowInfo);
   const classNamesToIgnore = [
```
tomer-openfin: maybe refer to those as blacklistedExternalWins or something like that? also, considering this list is probably going to get bigger, think about moving these string consts to their own file...

src/browser/utils.ts
```diff
@@ -74,20 +74,41 @@ function clamp(num: number, min: number = 0, max: number = Number.MAX_SAFE_INTEG
 }
 
 /*
-  Extends raw native window info.
+  Parses raw native window info object and returns a "clean" info object.
 */
-export function extendNativeWindowInfo(rawNativeWindowInfo: Shapes.RawNativeWindowInfo): Shapes.NativeWindowInfo {
+export function parseRawNativeWindowInfo(
```
tomer-openfin: alternative names to consider - getNativeWinInfoFromRawData / extractNativeWinInfoFromRawData

src/browser/api/external_window.ts
whyn07m3: 1. blacklistedExternalWins: "blacklisted" sounds a bit harsh, these external windows have done nothing wrong :) i'd prefer to keep something along ignore, omit, irrelevant, etc...
2. I don't think (hope) this list won't get much bigger, at least for Win10. Maybe Win7 might add a few here. Once it starts getting out of control, I agree, we should move them elsewhere then.

src/browser/utils.ts
whyn07m3: Not sure..., the thing is, props will be modified, removed or deleted. `get` is kinda there but not really. `extract` I don't think is a better word.

whyn07m3:
@tomer-openfin if looks good, please approve

whyn07m3:
@datamadic @tomer-openfin Ready for another review.
Note, one test for external windows started failing (`
fin.System 'external-window-created' event`), because it's relying on the information that's actually being taken out here. I asked @aziz512 to update the test.

PR #801, opened 2019-05-30
RUN-5235 fixed "blurred" event for external windows

whyn07m3:
#### Description of Change
This fixes external window's "blurred" event.
Prior to these changes, we relied on the injection to help us trigger "blurred" event. But, because "blur" from the injection appears to be broken and not wanting to rely on the injection, here we will use "focus" win event hook with some state tracking to implement "blurred" event for native windows.

#### Checklist

#### Release Notes
N/A


whyn07m3:
ah, got conflicts now

PR #805, opened 2019-05-31
Import to require

pbaize:
#### Description of Change
Changes import of js files from import to require as needed by typescript 3 upgrade

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: na tech debt


whyn07m3:
There's something about the failed tests that's concerning me :)

PR #807, opened 2019-06-03
[RUN-5033] Windowing API calls nack if size constraints would be violated

MichaelMCoates:
#### Description of Change
This PR modifies Windowing APIs that resize windows. If a Windowing API call attempts to resize a window outside of its `maxHeight`/`maxWidth`/`minHeight`/`minWidth` constraints, the function will nack. 

Affected functions:
`Window.animate`
`Window.resizeBy`
`Window.resizeTo`
`Window.setBounds`

**This PR will change the behavior of these API calls. Previously, the windows would resize up to their constraints, and the functions would ack. Now, we prevent all resizing, and nack instead**

The following tests will need to have a Max Version set @whyn07m3 :
https://testing-dashboard.openfin.co/#/app/tests/55897c432517bd926e4a257b/edit
https://testing-dashboard.openfin.co/#/app/tests/55897c432517bd926e4a257c/edit

The following tests will need to have a Min Version set (once this PR is merged and we've got a build number):
https://testing-dashboard.openfin.co/#/app/tests/5cf15765cb360141a7dfe572/edit
https://testing-dashboard.openfin.co/#/app/tests/5cf1515ecb360141a7dfe571/edit
https://testing-dashboard.openfin.co/#/app/tests/5ce46a6bcb360141a7dfe43a/edit
https://testing-dashboard.openfin.co/#/app/tests/5ce44742cb360141a7dfe42d/edit
https://testing-dashboard.openfin.co/#/app/tests/5ce441cecb360141a7dfe427/edit
https://testing-dashboard.openfin.co/#/app/tests/5ce44129cb360141a7dfe426/edit
https://testing-dashboard.openfin.co/#/app/tests/5ce31e0fcb360141a7dfe411/edit

#### Checklist

- [X] `npm test` passes
- [X] automated tests are [changed or added](https://testing-dashboard.openfin.co/#/app/dashboard)
- [?] manual tests are [changed or added](https://github.com/openfin/test_apps)
- [?] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter)

#### Release Notes
Windowing APIs that affect window size will now nack if that resize action would violate the window's size constraints.

Notes:

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1609,6 +1609,8 @@ function areNewBoundsWithinConstraints(options, width, height) {
 
     const acceptableWidth = (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
     const acceptableHeight = (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
+
+    // Check what the new aspect ratio would be at the proposed width/height. Precise to two decimal places.
     const roundedProposedRatio = Math.round(100 * (width / height)) / 100;
     const roundedAspectRatio = Math.round(100 * aspectRatio) / 100;
 
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1595,15 +1595,15 @@ function areNewBoundsWithinConstraints(options, width, height) {
         aspectRatio
     } = options;
 
-    if (width === undefined && height === undefined) {
+    if (typeof width !== 'number' && typeof height !== 'number') {
         return true;
     }
 
-    if (height === undefined) {
+    if (typeof height !== 'number') {
         return (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
     }
 
-    if (width === undefined) {
+    if (typeof width !== 'number') {
         return (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
     }
 
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1078,7 +1078,6 @@ Window.animate = function(identity, transitions, options = {}, callback = () => 
 
     if (!size) {
         animations.getAnimationHandler().add(browserWindow, animationMeta, animationTween, callback, errorCallback);
-        callback();
         return;
     }
 
@@ -1101,7 +1100,6 @@ Window.animate = function(identity, transitions, options = {}, callback = () => 
 
     if (newBoundsWithinConstraints) {
         animations.getAnimationHandler().add(browserWindow, animationMeta, animationTween, callback, errorCallback);
-        callback();
     } else {
         errorCallback(new Error(`Proposed window bounds violate size constraints for uuid: ${identity.uuid} name: ${identity.name}`));
     }
@@ -1597,8 +1595,20 @@ function areNewBoundsWithinConstraints(options, width, height) {
         aspectRatio
     } = options;
 
-    const acceptableWidth = width >= minWidth && width <= maxWidth;
-    const acceptableHeight = height >= minHeight && height <= maxHeight;
+    if (width === undefined && height === undefined) {
+        return true;
+    }
+
+    if (height === undefined) {
+        return (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
+    }
+
+    if (width === undefined) {
+        return (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
+    }
+
+    const acceptableWidth = (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
+    const acceptableHeight = (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
     const roundedProposedRatio = Math.round(100 * (width / height)) / 100;
     const roundedAspectRatio = Math.round(100 * aspectRatio) / 100;
 
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1558,24 +1558,69 @@ Window.removeEventListener = function(identity, type, listener) {
     ofEvents.removeListener(route.window(type, browserWindow.id), listener);
 };
 
+function areNewBoundsWithinConstraints(options, width, height) {
+    const {
+        minWidth,
+        minHeight,
+        maxWidth,
+        maxHeight,
+        aspectRatio
+    } = options;
 
-Window.resizeBy = function(identity, deltaWidth, deltaHeight, anchor) {
+    const acceptableWidth = width >= minWidth && width <= maxWidth;
+    const acceptableHeight = height >= minHeight && height <= maxHeight;
+    const roundedProposedRatio = Math.round(100 * (width / height)) / 100;
+    const roundedAspectRatio = Math.round(100 * aspectRatio) / 100;
+
+    return acceptableWidth && acceptableHeight && (aspectRatio <= 0 || roundedProposedRatio === roundedAspectRatio);
+}
+
+Window.resizeBy = function(identity, deltaWidth, deltaHeight, anchor, callback, errorCallback) {
     const browserWindow = getElectronBrowserWindow(identity);
     const opts = { anchor, deltaHeight, deltaWidth };
     if (!browserWindow) {
         return;
     }
-    NativeWindow.resizeBy(browserWindow, opts);
+
+    if (!('_options' in browserWindow)) {
+        errorCallback(new Error(`No window options present for uuid: ${identity.uuid} name: ${identity.name}`));
+        return;
+    }
+
+    const newWidth = browserWindow._options.width + deltaWidth;
+    const newHeight = browserWindow._options.height + deltaHeight;
+
+    const newBoundsAcceptable = areNewBoundsWithinConstraints(browserWindow._options, newWidth, newHeight);
+
+    if (newBoundsAcceptable) {
+        NativeWindow.resizeBy(browserWindow, opts);
+        callback({ success: true });
+    } else {
+        errorCallback(new Error(`Proposed window bounds violate size constraints for uuid: ${identity.uuid} name: ${identity.name}`));
+    }
 };
 
 
-Window.resizeTo = function(identity, width, height, anchor) {
+Window.resizeTo = function(identity, width, height, anchor, callback, errorCallback) {
     const browserWindow = getElectronBrowserWindow(identity);
     const opts = { anchor, height, width };
     if (!browserWindow) {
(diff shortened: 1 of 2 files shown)


MichaelMCoates:
PR Ready for review. I could use help with cleanup; I've got a lot of code duplication but not sure the best way to handle it.

src/browser/api/window.js
```diff
@@ -1074,7 +1074,35 @@ Window.animate = function(identity, transitions, options = {}, callback = () =>
         animationMeta.interrupt = true;
     }
 
-    animations.getAnimationHandler().add(browserWindow, animationMeta, animationTween, callback, errorCallback);
+    const { size } = transitions;
+
+    if (!size) {
+        animations.getAnimationHandler().add(browserWindow, animationMeta, animationTween, callback, errorCallback);
```
tomer-openfin: what's Tween?

src/browser/api/window.js
```diff
@@ -1558,24 +1586,81 @@ Window.removeEventListener = function(identity, type, listener) {
...
+        aspectRatio
+    } = options;
+
+    if (width === undefined && height === undefined) {
+        return true;
+    }
 
-Window.resizeBy = function(identity, deltaWidth, deltaHeight, anchor) {
+    if (height === undefined) {
+        return (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
```
tomer-openfin: how about maxWidth\Height <=0 here? I'm guessing users who want to remove max/min limitation would set it's value to 0.

src/browser/api/window.js
```diff
@@ -1558,24 +1586,81 @@ Window.removeEventListener = function(identity, type, listener) {
...
+    }
+
+    if (width === undefined) {
+        return (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
+    }
+
+    const acceptableWidth = (width >= minWidth) && (maxWidth === -1 || width <= maxWidth);
+    const acceptableHeight = (height >= minHeight) && (maxHeight === -1 || height <= maxHeight);
+    const roundedProposedRatio = Math.round(100 * (width / height)) / 100;
```
tomer-openfin: what's the motivation for rounding here? get a 2 decimal point precision? 
I'm not a big fan of comments, but I think this bit is technical enough to justify including some help for future readers

src/browser/api/window.js
```diff
@@ -1558,24 +1586,81 @@ Window.removeEventListener = function(identity, type, listener) {
...
+
+Window.resizeBy = function(identity, deltaWidth, deltaHeight, anchor, callback, errorCallback) {
     const browserWindow = getElectronBrowserWindow(identity);
     const opts = { anchor, deltaHeight, deltaWidth };
     if (!browserWindow) {
         return;
     }
-    NativeWindow.resizeBy(browserWindow, opts);
+
+    if (!('_options' in browserWindow)) {
```
tomer-openfin: Does anyone know if the middleware is testing for this?

tomer-openfin:
Question - in the following scenario: 
`Two windows are grouped and snapped. One of them has a minHeight. Someone is setting bounds on one of these windows to a lower value than the minHeight.`

For both the current implementation and ideally -

- would the call just nack and no change will occur *or* would the call ack with just 1 window changing? or a 3rd option?

- does it matter if the getBounds call is executed from the window with the minHeight or not?

whyn07m3:
@MichaelMCoates please, remind me to set max version for automated tests once this is merged and has a build number

whyn07m3:
@MichaelMCoates
image
the link above is not pointing to added or changed tests, please list the tests that were added or changed

src/browser/api/window.js
whyn07m3: Btw, this is being done a few times, good candidate for DRYing

src/browser/api/window.js
```diff
@@ -1558,24 +1586,81 @@ Window.removeEventListener = function(identity, type, listener) {
...
+        NativeWindow.resizeBy(browserWindow, opts);
+        callback();
+    } else {
+        errorCallback(new Error(`Proposed window bounds violate size constraints for uuid: ${identity.uuid} name: ${identity.name}`));
+    }
 };
 
 
-Window.resizeTo = function(identity, width, height, anchor) {
+Window.resizeTo = function(identity, width, height, anchor, callback, errorCallback) {
```
whyn07m3: Since these functions are synchronous, and I think we take care of this case, where we can simply throw an error instead of calling an error callback... not sure what's best.

MichaelMCoates:
> @MichaelMCoates
> image
> the link above is not pointing to added or changed tests, please list the tests that were added or changed. Same for other items in the list

Thanks, didn't realize that private tests were visible to others! Added.

src/browser/api/window.js
MichaelMCoates: https://cdn.openfin.co/jsdocs/stable/tutorial-window.animate.html

Different movement styles, essentially.

src/browser/api/window.js
dhamberlin: Shoutout to my man `Number.toFixed`. 

Some pros/cons vs this method: https://stackoverflow.com/a/12830454
tl;dr: returns a string, so you probably need to change it back to a number, but it handles rounding better in some cases.

^ this is probably a neckbeard nitpick and either version is probably fine for our purposes.

src/browser/api/window.js
MichaelMCoates: Anybody got info on what happens when you set maxWidth/Height to 0 vs -1? Seems like most of the code is expecting -1.

pbaize:
How about making a validation decorator function that just decorates the methods rather than changing each of them?

src/browser/api/window.js
pbaize: https://appoji.jira.com/browse/RUN-5019

src/browser/api/window.js
MichaelMCoates: Yeah, 2 decimal point precision. Ran a bunch of tests to see how the `aspectRatio` calculation changes depending on width/height, and it seems that our current manual resizing logic has 2-decimal precision. Is this too precise? Not precise enough? Adding comment.

src/browser/api/window.js
```diff
@@ -1558,24 +1586,81 @@ Window.removeEventListener = function(identity, type, listener) {
...
+function areNewBoundsWithinConstraints(options, width, height) {
+    const {
+        minWidth,
+        minHeight,
+        maxWidth,
+        maxHeight,
+        aspectRatio
+    } = options;
+
+    if (width === undefined && height === undefined) {
```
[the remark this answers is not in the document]
MichaelMCoates: Good call, changing.

src/browser/api/window.js
MichaelMCoates: With that in mind, should I leave as is then?

src/browser/api/window.js
MichaelMCoates: @dhamberlin oooooooo really cool, but looking at some of the comments underneath, seems like `Number.toFixed` also has rounding weirdness sometimes. This one in particular caught my eye, what do you think?:

> There is no advantage of using toFixed instead of Math.round; toFixed leads to quite the same rounding problems (try them with 5.555 and 1.005), but is like 500x (no kidding) slower than Math.round – Pierre May 20 '14 at 16:49

dhamberlin: Good looking out. I like `toFixed` in that its intent is clear (re: Tomer's comment). But this is one of the places in our code base where speed at this scale is liable to actually matter. Haven't benchmarked, but if `Math.round` is faster, we should probably use that.

dhamberlin: Follow up cause talking perf without testing squicks me out:

Ran some tests and this Pierre fellow is right. My results were a little less dramatic than 500x, but `toFixed` was decidedly slower (took about 2-3 times as long as `Math.round(100 * n) / 100` for most of my test runs)

src/browser/api/window.js
tomer-openfin: as I see it - some users will try passing falsy values in an attempt to cancel a min/max, let's save them the trip to the docs.

src/browser/api/window.js
MichaelMCoates: I was following the convention of the rest of the file, maybe we follow the convention in this PR and move it to error throwing in a separate PR?

MichaelMCoates:
> Question - in the following scenario:
> `Two windows are grouped and snapped. One of them has a minHeight. Someone is setting bounds on one of these windows to a lower value than the minHeight.`
> 
> For both the current implementation and ideally -
> 
> * would the call just nack and no change will occur _or_ would the call ack with just 1 window changing? or a 3rd option?
> * does it matter if the getBounds call is executed from the window with the minHeight or not?

Just tested on my machine. Created two windows, snapped them together, set a `maxHeight` of 600 pixels on one window (Window A), and tried to resize past 600px from both windows programmatically.

**9.61.38.43**
-> Resizing Window A: Window A resizes up to its constraint. Window B resizes up to the height set in `resizeTo`. Window B's width resizes strangely.
-> Resizing Window B: Window A resizes up to its constraint. Window B resizes to the bounds set in `resizeTo`

**11.69.42.19**
It seems that all resizing APIs don't do anything. The windows don't move, and there are no errors. 

**This PR**
Same as `11.69.42.19`. Because we hijack all move/resize API calls from grouped windows, we don't hit any of my code.

How should we handle this?

MichaelMCoates:
Looping in @datamadic @tgoc99 . Can you take a look at the above please?

tgoc99:
Looks good... From a grouped move perspective, it will need to be handled elsewhere.

tomer-openfin:
Thanks for running these tests! 

> **11.69.42.19**
> It seems that all resizing APIs don't do anything. The windows don't move, and there are no errors (which is bad, right?).
> 
> **This PR**
> Same as `11.69.42.19`. Because we hijack all move/resize API calls from grouped windows, we don't hit any of my code.

Does this probably deserve another ticket @datamadic?

tgoc99:
> Thanks for running these tests!
> 
> > **11.69.42.19**
> > It seems that all resizing APIs don't do anything. The windows don't move, and there are no errors (which is bad, right?).
> > **This PR**
> > Same as `11.69.42.19`. Because we hijack all move/resize API calls from grouped windows, we don't hit any of my code.
> 
> Does this deserve another ticket @datamadic?

Will be fixed with the new native / external window grouping PR.

PR #810, opened 2019-06-04
RUN-5333 native window injection heartbeat and reconnection

whyn07m3:
⚠️⏳ Pending changes to native window agent

#### Description of Change
This PR adds a heartbeat and reconnection / re-subscription logic to the injection bus.

#### Checklist

#### Release Notes
N/A

### DIFF
--- a/src/browser/transports/injection_bus.ts
+++ b/src/browser/transports/injection_bus.ts
@@ -46,10 +46,10 @@ interface NackMessage extends MessageBase {
 interface SendMessage {
   action: string;
   payload: {
-    data: {
+    data?: {
       [nativeId: string]: string[] | { userMovement: boolean; };
     },
-    type: string;
+    type?: string;
   };
 }
 
@@ -59,12 +59,17 @@ interface PendingRequest {
 }
 
 export default class NativeWindowInjectionBus extends EventEmitter {
+  private _connected: boolean; // Indicates whether the core is connected to the message window
+  private _listenerAck: (message: MessageBase) => void; // Listener called for ack/nack
+  private _listenerBroadcastMsg: (message: MessageBase) => void; // Listener called for broadcast messages
+  private _listenerHeartbeat: (message: MessageBase) => void; // Hearbeat listener
+  private _messageListener: (sender: number, rawMessage: string) => void; // Main message window listener
   private _meshUuid: string; // ID of core instance
-  private _messageListener: (sender: number, rawMessage: string) => void;
   private _nativeId: string; // HWND of the external window
   private _pendingRequests: Map<string, PendingRequest>;
   private _pid: number; // process ID of the external window
-  private _senderId: string; // ID of the injected DLL
+  private _previousDllSequence: number; // Last recorded sequence number from the message window
+  private _senderId: string; // ID of the message window of the injected process
 
   constructor(params: ConstructorParams) {
     super();
@@ -75,11 +80,9 @@ export default class NativeWindowInjectionBus extends EventEmitter {
     this._pendingRequests = new Map();
     this._pid = pid;
 
-    // Subscribe to all events
-    this.send({
-      action: 'window/subscription/request',
-      payload: { data: { [this._nativeId]: ['*'] }, type: 'set' }
-    });
+    this.setupHeartbeat();
+    this.setupListenerAck();
+    this.setupListenerBroadcastMsg();
 
     // Listen to messages from the transport and
     // forward broadcast messages locally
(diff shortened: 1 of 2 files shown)


src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
+        this._listenerAck(parsedMessage);
+      } else {
+        this._listenerBroadcastMsg(parsedMessage);
+      }
+    };
 
-        isNack
-          ? pendingRequest.reject((<NackMessage>parsedMessage).payload.reason)
-          : pendingRequest.resolve();
+    copyDataTransport.on('message', this._messageListener);
```
datamadic: Should this have a corresponding `removeListener` or this lives for the life of the transport?

src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
+
+    this.subscribe();
+  }
 
-        this._senderId = senderId;
-        this._pendingRequests.delete(messageId);
+  // Setup heartbeat
+  private setupHeartbeat() {
+    this._connected = false;
+    this._previousDllSequence = -1;
```
datamadic: :neckbeard: Why not call this the same thing as what comes of the message (`const { sequence } = message;`)
:neckbeard: :neckbeard: `sequence` implies multiple, perhaps `sequenceNumber` is better

src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
 
-        return;
+    this._listenerHeartbeat = (message: MessageBase): void => {
+      const { sequence } = message;
+
+      if (this._previousDllSequence + 1 !== sequence) {
```
datamadic: 1. After this is off once will it ever self correct or just log that it was missed over and over?
2. Should it raise and event or tell the other party that it was missed?
3. Is this ever meant to be user facing or is this a protocol level thing?

src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
+          + `previous sequence ${this._previousDllSequence}, `
+          + `current sequence: ${sequence}`;
+        writeToLog('info', msg);
+      }
+
+      this._previousDllSequence = sequence;
+    };
+
+    // Ping message window every second and update current connection status
+    setInterval(() => {
```
datamadic: This is not tied to the lifetime of a window? If so the interval needs to be cleaned up or its a leak

src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
+        this._connected = false;
+      });
+    }, 1000);
+  }
+
+  // Setup listener responsible for parsing acks
+  private setupListenerAck() {
+    this._listenerAck = (message: MessageBase): void => {
+      const { messageId, senderId } = message;
+      const isNack = message.action.includes('error');
```
datamadic: :neckbeard: perhaps make this a enum to be explicit and to able to handle diff types of errors w/o having to fish through strings.

src/browser/transports/injection_bus.ts
```diff
@@ -88,28 +91,84 @@ export default class NativeWindowInjectionBus extends EventEmitter {
...
+      this._senderId = senderId;
+      this._pendingRequests.delete(messageId);
+    };
+  }
+
+  // Setup listener responsible for parsing broadcast messages
+  private setupListenerBroadcastMsg() {
+    this._listenerBroadcastMsg = (message: MessageBase): void => {
+      if (!(<BroadcastMessage>message).payload || !(<BroadcastMessage>message).payload.data) {
+        writeToLog('info', `[NWI] Injection event without payload: ${JSON.stringify(message)}`);
```
datamadic: Is this actionable? Should we make an event out of this?

src/browser/transports/injection_bus.ts
whyn07m3: It's [here](https://github.com/HadoukenIO/core/blob/826c383cab220ca7e8feb8f8a289698093cac7e6/src/browser/transports/injection_bus.ts#L238)

src/browser/transports/injection_bus.ts
whyn07m3: 1. There are two types of sequences: sequence number coming from the core, and sequence number coming from the message window. Wanted to be more explicit to possibly avoid confusion
2. Ok for `sequenceNumber `

src/browser/transports/injection_bus.ts
whyn07m3: This is a protocol thing for now. Added this minimal logic here, because we still don't know what the behavior in this case should be. Technically, we shouldn't have any missed message as the sender knows when something fails to be send synchronously. This is just initial check, which will only log. Once and if the real-world problems start to arise for this case, we'll have to figure out how we will handle them.

src/browser/transports/injection_bus.ts
whyn07m3: Great catch 👍

src/browser/transports/injection_bus.ts
whyn07m3: Hmmm.. not sure it would work well for this case because of [the protocol](https://hackmd.io/z1ZaT6WcQlKMjweiycEP_w?view#ltnamespacegtlttopicgterror) 🤷‍♂

src/browser/transports/injection_bus.ts
whyn07m3: This was added initially and left there while we are ironing out parts of the communication protocol with the injection. I would like to  eventually remove it, this is here for now to make sure it doesn't throw.

src/browser/transports/injection_bus.ts
datamadic: Doesn't that say Int already or am I looking at the wrong thing? `code | <Integer> | The error code`

whyn07m3: It decides on whether it's an ack by parsing a part of a string vs looking at the property 🤷‍♂

src/browser/transports/injection_bus.ts
whyn07m3: added re-resubscription in this case

datamadic:
lets put this one on ice for a bit, if we circle back to the injection piece we can revisit

PR #811, opened 2019-06-05
RUN-5253 change export type to es modules

pbaize:
#### Description of Change
Changes exports in the core from using module.export to named exports. Required for typescript 3 upgrade

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: n/a

### DIFF
--- a/src/browser/api_protocol/api_handlers/application.js
+++ b/src/browser/api_protocol/api_handlers/application.js
@@ -32,7 +32,7 @@ let successAck = {
     success: true
 };
 
-module.exports.applicationApiMap = {
+export const applicationApiMap = {
     'close-application': closeApplication,
     'create-application': createApplication,
     'create-child-window': createChildWindow,
@@ -66,7 +66,7 @@ module.exports.applicationApiMap = {
     'wait-for-hung-application': waitForHungApplication
 };
 
-module.exports.init = function() {
+export const init = function() {
     apiProtocolBase.registerActionMap(module.exports.applicationApiMap, 'Application');
 };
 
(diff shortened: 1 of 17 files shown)


src/browser/deferred.ts
```diff
@@ -3,7 +3,7 @@ import { BrowserWindow as BrowserWindowOF } from '../shapes';
 import { toSafeInt } from '../common/safe_int';
 import { clipBounds } from './utils';
 
-export { handleMove };
+export default { handleMove };
```
whyn07m3: we can't remove this, can we?

pbaize: how so?

whyn07m3: Now, we do `import { handleMove } from './deferred';`. But we already `export function handleMove(` export this function. So if we remove this default export of a map, the imports won't change. It looks like an extra line to me, hope I'm not missing something

pbaize: It might be extra, left it in for paranoia. IMO it's never a bad thing to provide a default export of all named exports

whyn07m3: I'm 100% fine if you know it's good to have it there. I'm not okay if paranoia wrote this line of code :)

src/browser/api_protocol/api_handlers/application.js
```diff
@@ -66,7 +66,7 @@ module.exports.applicationApiMap = {
     'wait-for-hung-application': waitForHungApplication
 };
 
-module.exports.init = function() {
+export const init = function() {
     apiProtocolBase.registerActionMap(module.exports.applicationApiMap, 'Application');
```
MichaelMCoates: `module.exports` left over.

PR #813, opened 2019-06-05
fixed ts external window

whyn07m3:
#### Description of Change
This updates TS types for external windows due to recent main TS updates

#### Checklist

#### Release Notes
N/A

### DIFF
--- a/src/electron.d.ts
+++ b/src/electron.d.ts
@@ -3842,7 +3842,7 @@ declare namespace Electron {
     /**
      * The window's bounds.
      */
-    Bounds: Rectangle;
+    bounds: Rectangle;
     /**
      * The className of the window.
      */
--- a/src/shapes.ts
+++ b/src/shapes.ts
@@ -1,6 +1,6 @@
 
 import { PortInfo } from './browser/port_discovery';
-import { BrowserWindow as BrowserWindowElectron } from 'electron';
+import { BrowserWindow as BrowserWindowElectron, NativeWindowInfo } from 'electron';
 import { ERROR_BOX_TYPES } from './common/errors';
 import { AnchorType } from '../js-adapter/src/shapes';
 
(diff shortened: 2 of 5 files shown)


src/electron.d.ts
```diff
@@ -3842,7 +3842,7 @@ declare namespace Electron {
     /**
      * The window's bounds.
      */
-    Bounds: Rectangle;
```
whyn07m3: @pbaize it looks like a mistake, needs to be lowercase

pbaize: we should fix it in the runtime docs then, not here.

src/shapes.ts
```diff
@@ -451,21 +451,7 @@ export interface ExternalWindow extends BrowserWindowElectron {
...
-    id: string;
-    maximized: boolean;
-    minimized: boolean;
-    process: ProcessInfo;
-    title: string;
-    visible: boolean;
-}
-
-export interface NativeWindowInfo extends RawNativeWindowInfo {
+export interface NativeWindowInfo extends NativeWindowInfo {
```
dhamberlin: Seems odd to me to use the same name as the import from electron here. Maybe rename the electron import to something like `NativeWindowInfoElectron`(as we do for the `BrowserWindow` import).

This is a nitpick though and shouldn't really affect anything beyond aesthetics, so I'll go ahead and approve. But seems like something that might cause some small amount of future confusion.

src/electron.d.ts
whyn07m3: and there too. let's keep this here for now for it can work, otherwise we need to wait and now it won't work without this change

pbaize: feel free to generate types from a feature branch of the runtime

src/shapes.ts
whyn07m3: will change now

whyn07m3:
Gonna wait for test results to get in, then maybe merge right after

pbaize:
Can we make sure it doesn't break the other ts PR?

whyn07m3:
@pbaize what's the other TS PR? if you are referring to mine, then yeah it should work as I tested it together at the time of adding changes

PR #819, opened 2019-06-10
RUN-5060: reverting to always ack

tomer-openfin:
#### Description of Change
Reverting navigate behavior to ACKing synchronously no matter what happens.


tomer-openfin:
@HarsimranSingh done
@rdepena sent Wenjun the asar for testing. 
The event listeners are still there, but since we now ack immediately we're not waiting for them to fire. do you want me to remove them altogether from this pr?

rdepena:
Not required for v11

PR #821, opened 2019-06-11
RUN-5233 disabled external window frame

whyn07m3:
#### Description of Change
This change is for enabling external window's frame when all its requestors are gone.

#### Checklist

#### Release Notes
N/A

### DIFF
--- a/src/browser/api/external_window.ts
+++ b/src/browser/api/external_window.ts
@@ -11,12 +11,17 @@ import InjectionBus from '../transports/injection_bus';
 import ofEvents from '../of_events';
 import route from '../../common/route';
 import WindowGroups, { GroupChangedEvent, GroupEvent } from '../window_groups';
+import SubscriptionManager from '../subscription_manager';
 
 electronApp.on('ready', () => {
   subToGlobalWinEventHooks();
 });
 
+const subscriptionManager = new SubscriptionManager();
+
+// Maps
 export const externalWindows = new Map<string, Shapes.ExternalWindow>();
+const disabledUserMovementRequestorCount = new Map<string, number>();
 const externalWindowEventAdapters = new Map<string, ExternalWindowEventAdapter>();
 const injectionBuses = new Map<string, InjectionBus>();
 const windowGroupUnSubscriptions = new Map<string, () => void>();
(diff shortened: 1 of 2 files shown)


src/browser/api/external_window.ts
```diff
@@ -39,18 +44,44 @@ export function closeExternalWindow(identity: Identity): void {
...
   await injectionBus.set({ userMovement: false });
   externalWindow.emit('user-movement-disabled');
+
+  // Register a subscription that will fire every time requesting identity closes
+  // and if all requestors for disabling frame are gone - enable the frame.
+  if (requestor) {
+    const key = getKey(externalWindow);
+    const subscriptionKey = `disable-external-window-user-movement-${key}`;
+    const onRequestorClose = async () => {
+      let requestorCount = disabledUserMovementRequestorCount.get(key) || 0;
```
tomer-openfin: requestorCount scares me... wouldn't it be more foolproof to push\pop requestors to/from an array (instead of a map) and just check to see if the array is empty onRequestorClose?

whyn07m3: We do need a map to do the lookup to check which window has requestors. Are you saying you'd prefer to have an array of requestors for a particular window, instead of an integer (count)?

tomer-openfin: yes. it's a matter of preference, so definitely ignore this if you disagree. in my experience, it's easy to get these counters out of sync (assuming other people will poke this code in the future), in oppose to a more deterministic list of identities. this would also remove the need to decrement from enableExternaWindowUserMovement.
I'd would've used Map<string, Identity[]>

src/browser/api/external_window.ts
```diff
@@ -39,18 +44,44 @@ export function closeExternalWindow(identity: Identity): void {
...
+      if (requestorCount > 1) {
+        disabledUserMovementRequestorCount.set(key, --requestorCount);
+      } else {
+        await enableExternaWindowUserMovement(target);
```
tomer-openfin: one other way of removing the decrement from enableExternaWindowUserMovement would be to call it here when `requestorCount === 1`

PR #848, opened 2019-06-27
RUN-5392 - group resize bugfix

tgoc99:
#### Description of Change
When RTv10 windows are tabbed by the Layout Service they can't be resized vertically from the top or the bottom.  The bottom is a bug that will affect not only tabbed windows.  There is a bug where when there are 3+ windows, the move will not get propagated to all of the windows.  This has been replicated with two windows on top of each other (same bounds), then a third window on any side of the window.  Try to resize either of the windows with identical bounds - in the direction towards or away from the third window and it will fail to resize.  

This is related to this [JIRA](https://appoji.jira.com/browse/RUN-5348) but only fixes half the issue.

#### Checklist

- [ x ] PR description included and stakeholders cc'd
- [ x ] `npm test` passes
- [ x ] automated tests are added (see unit test in PR)
- [ x ] PR title starts with the JIRA ticket [pull request process](https://github.com/openfin/Internal-Wiki/wiki/Pull-Request-Process)
- [ ] PR release notes describe the change in a way relevant to app-developers (need to fully fix problem first)

#### Release Notes

Notes: NA - need to fully fix problem first.  Probably more so a layouts release not but if we have a runtime one would be on the JIRA linked to above.

### DIFF
--- a/src/browser/rectangle.ts
+++ b/src/browser/rectangle.ts
@@ -7,8 +7,6 @@ type SharedBounds = {
     left: SideName;
 };
 type SharedBound = Array<SideName>;
-type BoundIdentifier = [Rectangle, SideName];
-type RectangleBaseKeys = 'x' | 'y' | 'width' | 'height';
 export type SharedBoundsList = Array<SharedBound>;
 type Graph = [number[], number[][]];
 export interface Opts {
@@ -53,17 +51,6 @@ export class Rectangle {
         return new Rectangle(x, y, width, height, new RectOptionsOpts(options));
     }
     public static BOUND_SHARE_THRESHOLD = 5;
-
-    private static readonly EDGE_CROSSINGS: SideName[][] = [
-        ['top', 'top'],
-        ['top', 'bottom'],
-        ['bottom', 'top'],
-        ['bottom', 'bottom'],
-        ['right', 'left'],
-        ['right', 'right'],
-        ['left', 'left'],
-        ['left', 'right']
-    ];
 
     public x: number;
     public y: number;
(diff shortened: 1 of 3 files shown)


src/browser/rectangle.ts
```diff
@@ -600,9 +589,11 @@ function propMoveThroughGraph (
     let movedRef = rects[refVertex];
 
     if (movedRef.hasIdenticalBounds( cachedBounds)) {
+        // This is the leader, move it to the proposed bounds
```
pbaize: Should we check if the new bounds violate size constraints or is that taken care of? @MichaelMCoates

pbaize: Seems to be good in this regard

src/browser/rectangle.ts
```diff
@@ -613,24 +604,24 @@ function propMoveThroughGraph (
...
-        e.forEach(([u, v]) => {
-            if (!visited.includes(v)) {
-                if (distances.get(v) === Infinity) {
+            e.forEach(([u, v]) => {
+                if (!visited.includes(v) && distances.get(v) === Infinity) {
                     toVisit.push(v);
                     distances.set(v, distances.get(u) + 1);
                     
-                    propMoveThroughGraph(rects, v, rects[refVertex], movedRef, visited);
+                    const visitedClone = JSON.parse(JSON.stringify(visited));
```
pbaize: :neckbeard: 
```suggestion
                    const visitedClone = [...visited];
```

tgoc99: :+1:

src/browser/rectangle.ts
```diff
@@ -355,12 +342,14 @@ export class Rectangle {
         return new Rectangle(this.x + delta.x, this.y + delta.y, this.width + delta.width, this.height + delta.height, this.opts);
     }
 
-    public move = (cachedBounds: RectangleBase, currentBounds: RectangleBase): Rectangle => {
+    public propogateMoveToThisRect = (leaderStartBounds: RectangleBase, proposedLeaderMove: RectangleBase): Rectangle => {
```
MichaelMCoates: *`propagate`

src/browser/rectangle.ts
MichaelMCoates: My size constraint violation stuff was only on API calls.

src/browser/rectangle.ts
tgoc99: :+1:

PR #869, opened 2019-07-22
Fix Channels API iFrame bug

aziz512:
The PR fixes a bug where if both channel provider creation and client connection happen in two different iframes of the same window, the provider promise resolves but the client promise never does.

#### Checklist
- [x] `npm test` passes

@rdepena @datamadic @dhamberlin 
#### Release Notes
This allows to create channels from iframe entities.

### DIFF
--- a/src/browser/core_state.ts
+++ b/src/browser/core_state.ts
@@ -125,7 +125,7 @@ export function getEntityInfo(identity: Shapes.Identity) {
 }
 
 export function getExternalOrOfWindowIdentity(identity: Shapes.Identity): Shapes.ProviderIdentity|undefined {
-    const { uuid, name } = identity;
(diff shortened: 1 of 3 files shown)


src/browser/core_state.ts
```diff
@@ -125,7 +125,7 @@ export function getEntityInfo(identity: Shapes.Identity) {
 }
 
 export function getExternalOrOfWindowIdentity(identity: Shapes.Identity): Shapes.ProviderIdentity|undefined {
-    const { uuid, name } = identity;
```
tgoc99: maybe rename to get identity & entity type?  Name no longer really applicable.

aziz512: Good idea. Renamed to getEntityIdentity as it still returns identity but not only for windows.

PR #873, opened 2019-07-25
RUN-5420: Keep reference to created context menu (#870)

CezaryKulakowski:
Without it context menu will be destroyed after first launch of
javascript's garbage collector even when menu is still displayed.

This is a backport of a fix from of-master.

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -10,6 +10,7 @@ let BrowserWindow = electron.BrowserWindow;
 let electronApp = electron.app;
 let Menu = electron.Menu;
 let nativeImage = electron.nativeImage;
+let currentContextMenu = null;
 
 // npm modules
 let _ = require('underscore');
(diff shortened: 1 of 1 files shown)


src/browser/api/window.js
```diff
@@ -1774,9 +1775,13 @@ Window.showMenu = function(identity, x, y, editable, hasSelectedText) {
         accelerator: 'CommandOrControl+Shift+I'
     });
 
-    const currentMenu = Menu.buildFromTemplate(menuTemplate);
-    currentMenu.popup(browserWindow, {
-        async: true
+    currentContextMenu = Menu.buildFromTemplate(menuTemplate);
+    currentContextMenu.popup({
+        window: browserWindow,
+        async: true,
```
HarsimranSingh: Noticing that `async` doesn't show up in https://github.com/HadoukenIO/core/blob/13.76.43/staging/src/electron.d.ts#L13267

Surprised it doesn't complain about this.

CezaryKulakowski: It was removed here:
https://github.com/electron/electron/commit/9c5acf623397331009876f163d22ad3abe02687d
so right now it has no effect. As Menu.popup takes a list as parameter you can pass additional elements but they will not be read. Additionaly the call was wrong as `browserWindow` was treated as a list of parameters and - as result - all parameters were set to default. Magic of javascript...

PR #882, opened 2019-08-05
Run 5463 nwi api security

dhamberlin:
#### Description of Change
Jira: https://appoji.jira.com/browse/RUN-5463

This change places `System.getAllExternalWindows` and `ExternalWindow.wrap` (sent to core as `register-native-external-window`) behind API security. I've opted to just place these two methods behind security in the interest of simplicity / ease of use:
- These are the two entry points to "doing stuff with native windows". If both are disallowed, it should, to the best of my analysis, be impossible to get info about or do anything to non-OpenFin windows.
- Current changeset uses the same exact flow as other restricted API methods, so we don't need extra explanation or code to make it work. (As opposed to an `ExternalWindow`-level boolean setting, which would require additional code changes and counterintuitively also control `System.getAllExternalWindows`)

@aziz512 (tagging you cause github won't let me assign you as a reviewer)

#### Checklist
- [x] `npm test` passes
- [ ] automated tests are [changed or added](https://testing-dashboard.openfin.co/#/app/dashboard) PENDING WILL UPDATE SHORTLY
- [ ] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter) PENDING WILL UPDATE SHORTLY
- [ ] Relevant changes to schemastore JSON schema submitted - PENDING

#### Release Notes

Notes: ExternalWindow functionality has been placed behind API security. To allow its use, `System.getAllExternalWindows` and `ExternalWindow.wrap` should be enabled via desktop owner settings or app manifest.

### DIFF
--- a/src/browser/api_protocol/api_handlers/external_window.ts
+++ b/src/browser/api_protocol/api_handlers/external_window.ts
@@ -25,7 +25,11 @@ export const ExternalWindowApiMap: APIHandlerMap = {
   'minimize-external-window': minimizeExternalWindow,
   'move-external-window-by': moveExternalWindowBy,
   'move-external-window': moveExternalWindow,
-  'register-native-external-window': registerNativeExternalWindow,
+  'register-native-external-window': {
+    apiFunc: registerNativeExternalWindow,
+    apiPath: '.registerNativeExternalWindow',
+    defaultPermission: false
+  },
   'resize-external-window-by': resizeExternalWindowBy,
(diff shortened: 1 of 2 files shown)


whyn07m3:
so this will require adding new API permission to the test runner, correct?

dhamberlin:
> so this will require adding new API permission to the test runner, correct?

Correct. Gonna work on that, as well as adding a couple tests to make sure it's working properly. (at least for the window options case)

PR #885, opened 2019-08-06
RUN-5206 Set up webcontents eventing

pbaize:
#### Description of Change
Sets up webcontents event hooks for browserview

#### Checklist

- [x] `npm test` passes
- [ ] Link to new tests added

#### Release Notes

Notes: n/a experimental

### DIFF
--- a/src/browser/api/webcontents.ts
+++ b/src/browser/api/webcontents.ts
@@ -1,4 +1,60 @@
 import * as url from 'url';
+import { Identity } from '../../shapes';
+import of_events from '../of_events';
+import { WindowRoute } from '../../common/route';
+
+export function hookWebContentsEvents(webContents: Electron.WebContents, { uuid, name }: Identity, topic: string, route: WindowRoute) {
+    webContents.on('did-get-response-details', (e,
+        status,
+        newUrl,
+        originalUrl,
+        httpResponseCode,
+        requestMethod,
+        referrer,
+        headers,
+        resourceType
+    ) => {
+        const type = 'resource-response-received';
+
+        const payload = {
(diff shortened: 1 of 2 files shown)


src/browser/api/webcontents.ts
```diff
@@ -1,9 +1,63 @@
...
+        originalUrl,
+        httpResponseCode,
+        requestMethod,
+        referrer,
+        headers,
+        resourceType
+    ) => {
+        const type = 'resource-response-received';
+
+        const payload = {
```
MichaelMCoates: Looks great, the only big difference I can see is that instead of `Object.assign`ing the payload, we're picking out specific items to include in the payload. On a quick glance over the code, it seems like this is fine, so I'm assuming you've considered this, yeah?

src/browser/api/webcontents.ts
```diff
@@ -1,9 +1,63 @@
...
+            topic,
+            type,
+            errorCode,
+            errorDescription,
+            validatedURL,
+            isMainFrame
+        };
+        ofEvents.emit(routeFunc(type, uuid, name), payload);
+    });
+    webContents.once('destroyed', () => {
```
MichaelMCoates: `close` has been changed to `destroyed`. I'm assuming this is `webContents` specific?

pbaize: yes

src/browser/api/webcontents.ts
pbaize: yep, this new format plays nicer with typescript and is more readable. This used to be set in a call that was modifying the payload

PR #894, opened 2019-08-12
RUN-5489   BrowserView name should be unique

aziz512:
#### Description of Change
Add a check of whether the name provided in options is already in use within the same application by either another browser view or a window.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: Limited BrowserView identities to unique names

@pbaize @rdepena @bschwinn

### DIFF
--- a/src/browser/api/browser_view.ts
+++ b/src/browser/api/browser_view.ts
@@ -25,7 +25,8 @@ export async function create(options: BrowserViewOpts) {
     // checking if the name-uuid combination is already in use
     const { uuid, name } = options;
     if (getWindowByUuidName(uuid, name) || getBrowserViewByIdentity({ uuid, name }) || getInfoByUuidFrame({ uuid, name })) {
-        throw new Error('Provided name-uuid combination is already in use');
+        throw new Error('Trying to create a BrowserView with name-uuid combination already in use - '
+            + JSON.stringify({ name, uuid }));
     }
 
     if (!options.target) {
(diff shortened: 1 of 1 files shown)


src/browser/api/browser_view.ts
```diff
@@ -19,6 +19,12 @@ export interface BrowserViewOpts extends BrowserViewCreationOptions {
 }
 
 export async function create(options: BrowserViewOpts) {
+    // checking if the name-uuid combination is already in use
+    const { uuid, name } = options;
+    if (getWindowByUuidName(uuid, name) || getBrowserViewByIdentity({ uuid, name })) {
```
pbaize: Should we check frames too?

aziz512: Done :+1:

src/browser/api/browser_view.ts
```diff
@@ -19,6 +22,12 @@ export interface BrowserViewOpts extends BrowserViewCreationOptions {
 }
 
 export async function create(options: BrowserViewOpts) {
+    // checking if the name-uuid combination is already in use
+    const { uuid, name } = options;
+    if (getWindowByUuidName(uuid, name) || getBrowserViewByIdentity({ uuid, name }) || getInfoByUuidFrame({ uuid, name })) {
+        throw new Error('Provided name-uuid combination is already in use');
```
pbaize: Can we make this message more inline with the window one? Let's also add what the uuid/name combination is in the message.

aziz512: :+1:

PR #895, opened 2019-08-13
[n/a] removed dead code that should not work

pbaize:
#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: n/a

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -1317,13 +1317,6 @@ Window.getOptions = function(identity) {
     }
 };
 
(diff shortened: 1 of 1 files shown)


pbaize:
> https://testing-dashboard.openfin.co/#/app/tests/5a9daa046a994a57faa5c9d9/edit
> 
> Looks like we have a couple of tests in the testing-dashboard using `Window.getParentApplication`. Have those been failing/how will they be affected/do they need to be removed?

The code for that is all adapter side. This code has no routing

PR #898, opened 2019-08-14
RUN-5484 Fix infinite message loop issue

licui3936:
This PR is to fix an infinite message loop issue using duplicate uuid in multi runtime environments. It causes browser process  and debug.log size increasing dramatically. 

Will add a test later. 
#### Checklist

- [x] `npm test` passes

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,7 +54,7 @@ import route from './src/common/route';
 
 import { createWillDownloadEventListener } from './src/browser/api/file_download';
 import duplicateUuidTransport from './src/browser/duplicate_uuid_delegation';
-import { deleteApp, argv } from './src/browser/core_state';
+import { deleteApp } from './src/browser/core_state';
 import { lockUuid } from './src/browser/uuid_availability';
 
 // locals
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -639,7 +639,16 @@ function launchApp(argo, startExternalAdapterServer) {
         if (uuid && !isRunning) {
             if (!lockUuid(uuid)) {
                 deleteApp(uuid);
-                duplicateUuidTransport.broadcast({ argv, uuid });
+                // We need to rebuild a new argv to have correct app info in it.
+                let newArgv = Object.keys(argo).map(key => {
+                    if (key === '_') {
```
pbaize: would this value ever be an array with a length other than 1? I believe it's all arguments that can't be split into flag/value pairs

rdepena:
Hey @licui3936 can we make this change against staging as well ?

index.js
licui3936: The value is array, but not name-value pair in array.

licui3936:
@rdepena should i close this one? just create PR for staging?

licui3936:
Create a PR for staging so i close this one.

PR #909, opened 2019-08-21
RUN-5507 Switch to using webcontents id rather than browserwindow id

pbaize:
#### Description of Change
Switch to using webcontents id rather than browserwindow id for core ids.

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: n/a pre-req for browser-view support

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ import socketServer from './src/browser/transports/socket_server';
 
 import { addPendingAuthRequests, createAuthUI } from './src/browser/authentication_delegate';
 let convertOptions = require('./src/browser/convert_options.js');
-let coreState = require('./src/browser/core_state.js');
+import * as coreState from './src/browser/core_state';
 let webRequestHandlers = require('./src/browser/web_request_handler.js');
 let errors = require('./src/common/errors.js');
 import ofEvents from './src/browser/of_events';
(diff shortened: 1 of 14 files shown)


index.js
```diff
@@ -68,8 +68,13 @@ const serverReadyPromise = new Promise((resolve) => {
...
 });
 
-app.on('child-window-created', function(parentId, childId, childOptions) {
-
+//Event either comes from the runtime or the core when registering an externalWindow.
+// Payload is a browserWindow id
+app.on('child-window-created', function(parentBwId, childBwId, childOptions) {
+    const parent = BrowserWindow.fromId(parentBwId);
+    const child = BrowserWindow.fromId(childBwId);
+    const parentId = parent.webContents.id;
```
MichaelMCoates: Do these need `undefined` checks?  `parent && parent.webContents && parent.webContents.id`?

pbaize: they may need a top level check for parent. That being said if it was ever undefined we'd be in a terrible place anyways so not sure what we should be doing if the check fails.

src/browser/core_state.ts
```diff
@@ -484,10 +484,15 @@ export function deleteApp(uuid: string): void {
     apps = apps.filter(app => app.uuid !== uuid);
 }
 
-export function getWindowOptionsById(id: number): Shapes.WindowOptions|boolean {
+export function getWindowOptionsById(id: number): Shapes.WindowOptions|false {
     const win = getWinById(id);
     return win && win.openfinWindow && win.openfinWindow._options;
 }
+export function getWindowOptionsByBrowserWindowId(id: number): Shapes.WindowOptions|false {
+    const win =  getWinList().find(win =>
```
MichaelMCoates: Nitpick: Maybe use a different variable to avoid confusion?

src/browser/navigation_validation.ts
```diff
@@ -52,8 +52,8 @@ export function navigationValidator(uuid: string, name: string, id: number) {
         const appObject = coreState.getAppObjByUuid(uuid);
         const appMetaInfo = coreState.appByUuid(uuid);
         const isMailTo = /^mailto:/i.test(url);
-        const allowed = isMailTo || validateNavigationRules(uuid, url, appMetaInfo.parentUuid, appObject._options) &&
-                                    isURLAllowed(url);
+        const allowed = isMailTo || (appObject && validateNavigationRules(uuid, url, appMetaInfo.parentUuid, appObject._options) &&
```
MichaelMCoates: Do we have to check `appMetaInfo` as well?

pbaize: I only added this check for typescript reasons, we don't currently test it and it builds.

PR #923, opened 2019-09-05
returning View's target with getInfo

tomer-openfin:
#### Description of Change
browserView.getInfo now also returns the target

- [x] `npm test` passes


tomer-openfin:
Does it make sense to you to return the target with getInfo? one alternative would be a specific getTarget endpoint.

pbaize:
View.getCurrentWindow ? Keep getInfo tied to we contents stuff?

tomer-openfin:
Closing this PR and opening a new one only with adapter changes to incorporate getCurrentWindow call

PR #939, opened 2019-09-16
Run 5493 external window id updates

dhamberlin:
#### Description of Change
Jira: https://appoji.jira.com/browse/RUN-5493

This PR makes a few changes to the ExternalWindow API:
- ExternalWindows can be registered by `nativeId` (nativeId === hwnd in Windows) OR uuid (if the uuid has already been assigned to an external process via e.g. `System.launchExternalProcess`
- `ExternalWindow.wrap` returns a uuid, name, and nativeId to caller, ensuring that render-side `ExternalWindow`s will always have all three, even if user wrapped the window by `uuid` or `nativeId` alone.
- ExternalWindow uuids are locked on registration, released on window close.
- nativeId is exposed on all ExternalWindow info objects, such as those returned by `System.getAllExternalWindows` as well as on 'external-window' events.
- uuid in included on all ExternalWindow info objects and external window events if a) the window has already been wrapped as an ExternalWindow or b) the pid matches a process that is being tracked by ProcessTracker (e.g. those launched by `launchExternalProcess`)

Related js-adapter PR: https://github.com/HadoukenIO/js-adapter/pull/341

#### Checklist

- [x] `npm test` passes
- [x] automated tests are [changed or added](https://testing-dashboard.openfin.co/#/app/dashboard) ( a couple private ones I will make public post-merge, and more to come)
- [x] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter) (more to come here, working on adding/updating tutorials now)
- [x] PR release notes describe the change in a way relevant to app-developers (included in js-adapter PR)

#### Release Notes

Notes:


src/browser/api/external_window.ts
```diff
@@ -693,7 +695,9 @@ function externalWindowCloseCleanup(externalWindow: Shapes.ExternalWindow): void
 
   externalWindow.emit('closed');
   externalWindow.removeAllListeners();
-  externalWindows.delete(nativeId);
+  externalWindows.delete(uuid);
+  nativeIdToUuid.delete(nativeId);
```
rdepena, line 707: We should guarantee a few of these things right, if we get an error on a preceding step we should still release the nativeUuid no?

dhamberlin: @rdepena Most of this cleanup seems relatively fail-proof. Wrapped the bits that might fail in a try catch.

src/browser/api/external_window.ts
```diff
@@ -532,10 +534,10 @@ function subscribeToWinEventHooks(externalWindow: Shapes.ExternalWindow): void {
 
 // Window grouping stub (makes external windows work with our original disabled frame group tracker)
 // Also some of the _options' values are needed in OpenFin Layouts
-function setAdditionalProperties(externalWindow: Shapes.ExternalWindow, properIdentity: Shapes.NativeWindowIdentity): Shapes.GroupWindow {
+function setAdditionalProperties(externalWindow: Shapes.ExternalWindow, requestedId: Shapes.NativeWindowIdentity): Shapes.GroupWindow {
```
licui3936: Feel like 'requestedId' is a string.  How about 'requestedIdentity' instead?

dhamberlin: Makes sense to me. Updated.

src/browser/api/external_window.ts
```diff
@@ -532,10 +534,13 @@ function subscribeToWinEventHooks(externalWindow: Shapes.ExternalWindow): void {
...
 // Also some of the _options' values are needed in OpenFin Layouts
-function setAdditionalProperties(externalWindow: Shapes.ExternalWindow, properIdentity: Shapes.NativeWindowIdentity): Shapes.GroupWindow {
+function setAdditionalProperties(
+  externalWindow: Shapes.ExternalWindow,
+  requestedIdentity: Shapes.NativeWindowIdentity
+): Shapes.GroupWindow {
   const { nativeId } = externalWindow;
-  const uuid = properIdentity.uuid || nativeId;
-  const name = properIdentity.name || nativeId;
+  const uuid = externalWindow.uuid || requestedIdentity.uuid || nativeId;
```
tomer-openfin: is it possible that the nativeId here is already taken as a uuid by some other window?

dhamberlin: Thanks for catching this. Actually not a problem because we ensure that we have a usable uuid before we call this function (line 273: `const uuid = identity.uuid || electronApp.generateGUID();`, so we'll try a user-provided uuid if included, otherwise use a new guid).

This does mean that these `||` are unnecessary. Removed them accordingly.

PR #945, opened 2019-09-17
RUN-5537: Always pass headers in onBeforeSendHeaders callback.

marekharanczyk:
New network service web request api implementation always reassigns
headers to whatever is passed in callback. Omitting headers when we did
not modify them ended up removing all previously assigned headers from
request.


pbaize:
Should this also go into develop?

PR #970, opened 2019-09-25
Bugfix - Call removeBrowserView on oldWin so that we can attach BrowserViews back and forth

MichaelMCoates:
#### Description of Change

Fixes a bug that prevents a BrowserView from attaching back to the window it was originally attached to. 

Bug Repro:
1. Create a BrowserView on Window A.
2. Attach the BV to Window B.
3. Attempt to attach the BV back to Window A.
Expected: BV Shows up on Window A.
Actual: BV stays on Window B.

To fix, we simply call removeBrowserView with the old window.

#### Checklist

- [X] `npm test` passes)

#### Release Notes

Notes:


pbaize:
Let's also move the target check above this section

MichaelMCoates:
> Let's also move the target check above this section

Yep, that's in a separate PR.

PR #976, opened 2019-10-02
[N/A] Hook up event propagation for all browserview-related events

MichaelMCoates:
#### Description of Change

This PR adds propagated events up to Application and System for BrowserView.

Tests:
https://testing-dashboard.openfin.co/#/app/sessions/api/completed/5da8b1ceee166d7a14d451bc

Corresponding js-adapter PR:
https://github.com/HadoukenIO/js-adapter/pull/357

@aziz512 trying to add you as reviewer but it won't let me for some reason.

#### Checklist

- [X] `npm test` passes
- [X] automated tests are [changed or added](https://testing-dashboard.openfin.co/#/app/dashboard)
- [X] relevant documentation is [changed or added](https://github.com/hadoukenio/js-adapter)
- [X] Link to new tests added

#### Release Notes

Notes:

### DIFF
--- a/src/browser/api/browser_view.ts
+++ b/src/browser/api/browser_view.ts
@@ -149,13 +149,6 @@ export async function attach(ofView: OfView, toIdentity: Identity) {
             target: toIdentity,
             previousTarget
         });
-        of_events.emit(route.window('view-attached', toIdentity.uuid, toIdentity.name), {
-            name: ofView.name,
-            uuid: ofView.uuid,
-            viewIdentity: {uuid: ofView.uuid, name: ofView.name},
-            target: toIdentity,
-            previousTarget
-        });
     }
 }
 export async function destroy (ofView: OfView) {
--- a/src/browser/api/browser_view.ts
+++ b/src/browser/api/browser_view.ts
@@ -94,9 +94,15 @@ export async function attach(ofView: OfView, toIdentity: Identity) {
             const oldWin = getWindowByUuidName(previousTarget.uuid, previousTarget.name);
             if (oldWin) {
                 oldWin.browserWindow.removeBrowserView(view);
-                of_events.emit(route.window('view-detached', previousTarget.uuid, previousTarget.name), {
-                    uuid: ofView.uuid,
+                of_events.emit(route.view('detached', ofView.uuid, ofView.name), {
                     name: ofView.name,
+                    uuid: ofView.uuid,
+                    target: toIdentity,
+                    previousTarget
+                });
+                of_events.emit(route.window('view-detached', previousTarget.uuid, previousTarget.name), {
+                    name: ofView.name,
+                    uuid: ofView.uuid,
                     viewIdentity: {uuid: ofView.uuid, name: ofView.name},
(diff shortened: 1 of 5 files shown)


src/browser/of_events.ts
```diff
@@ -33,6 +34,7 @@ class OFEvents extends EventEmitter {
         if (tokenizedRoute.length >= 2) {
             const [channel, topic] = tokenizedRoute;
             const uuid: string = (payload && payload.uuid) || tokenizedRoute[2] || '*';
+            const name: string|false = (payload && payload.name);
```
dhamberlin: Mostly irrelevant nitpick: will eval to `undefined` if there's no payload or no name.
`(!!payload && payload.name)` will evaluate to `false` if there's no payload, but I think unidentified is probably actually what we want in this case anyhow.

src/browser/api/browser_view.ts
```diff
@@ -103,11 +103,20 @@ export async function attach(ofView: OfView, toIdentity: Identity) {
...
+            of_events.emit(route.view('detached', ofView.uuid, ofView.name), {
+                name: ofView.name,
+                uuid: ofView.uuid,
+                target: toIdentity,
+                previousTarget
+            });
             of_events.emit(route.window('view-detached', previousTarget.uuid, previousTarget.name), {
+                name: ofView.name,
+                uuid: ofView.uuid,
                 viewIdentity: {uuid: ofView.uuid, name: ofView.name},
```
aziz512: Do we need the duplicate `viewIdentity` property here?

aziz512 (changes requested):

src/browser/api/browser_view.ts
MichaelMCoates: Yes, because this name/uuid gets stomped on in `of_events`. In general, I found that all the overwriting for event can get pretty dicey, so I wanted to make it nice and explicit.

src/browser/of_events.ts
MichaelMCoates: Agreed. Is there a case in which this could not work as expected?

dhamberlin: I don't think so. I would remove the `|false` typing, since this will never eval to `false`. But code looks good as is.

MichaelMCoates: Ah, right, good call, missed that. Changed.

aziz512 (dismissed):

dhamberlin (dismissed):
lgtm

pbaize (changes requested):
If we restrict views to targets in the same app do we still need the custom window logic?

MichaelMCoates:
The custom window logic is still necessary for `view-detached` because of the way the routing works in `of_events`.

MichaelMCoates:
After talking with Pierre and Tommy, changed the eventing scheme to the following:

- Windows can listen to `view-attached` and `view-detached`.
- Views can listen to `target-changed`. 
- Window `view` events don't propagate up to Application and System.
- All `view` events propagate up to Window, Application, and System.

The tests have been updated to reflect these changes.

PR #978, opened 2019-10-03
[RUN-5415] Add nameAlias check for external connections to sendChannelMessage

MichaelMCoates:
#### Description of Change

Add `nameAlias` check for external connections to `sendChannelMessage`.
Enables channel sub-addressing for external connections.

#### Checklist

- [X] `npm test` passes

#### Release Notes

Notes:

### DIFF
--- a/src/browser/api/channel.ts
+++ b/src/browser/api/channel.ts
@@ -181,6 +181,12 @@ export module Channel {
         const { uuid, name, payload: messagePayload, action: channelAction, providerIdentity } = payload;
         const intendedTargetIdentity = { uuid, name };
 
+        const messagingWindow = getEntityIdentity(identity);
+
+        if (messagingWindow && messagingWindow.isExternal && messagePayload && messagePayload.nameAlias) {
+            identity.name = messagePayload.nameAlias;
+        }
+
         const ackToSender = createAckToSender(identity, messageId, providerIdentity);
 
         sendToIdentity(intendedTargetIdentity, {


src/browser/api/channel.ts
```diff
@@ -181,6 +181,12 @@ export module Channel {
         const { uuid, name, payload: messagePayload, action: channelAction, providerIdentity } = payload;
         const intendedTargetIdentity = { uuid, name };
 
+        const messagingWindow = getEntityIdentity(identity);
+
```
licui3936, line 185: This getEntityIdentity function returns ProviderIdentity.  Maybe we need to change the name. It's confusing. 
Otherwise it's ok for me.

MichaelMCoates: @aziz512 what do you think? Seems like that function only returns {uuid, name, isExternal}. Do we give it a different return type?

dhamberlin: Maybe give it its own type? PotentiallyExternalIdentity feels awkward but would describe it.

dhamberlin (approved):
Agree with @licui3936's feedback re: type naming. Would be dope to get that into this PR. LGTM otherwise. I'll approve since that change isn't strictly in scope for this PR, but would be a solid nice-to-have.

src/browser/api/channel.ts
rdepena: This function is defined outside of this PR right, we should create a line item for this.

rdepena (approved):

MichaelMCoates:
JIRA created for the type change.

PR #986, opened 2019-10-07
Window refresh 2 named exports

pbaize:
#### Description of Change
Switch to named exports from window.js

#### Checklist

- [x] `npm test` passes

#### Release Notes

Notes: n/a internal

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ let minimist = require('minimist');
 // local modules
 let Application = require('./src/browser/api/application.js').Application;
 let System = require('./src/browser/api/system.js').System;
-import { Window } from './src/browser/api/window';
+import { create } from './src/browser/api/window';
 
 let apiProtocol = require('./src/browser/api_protocol');
 import socketServer from './src/browser/transports/socket_server';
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -21,7 +21,7 @@ let minimist = require('minimist');
 // local modules
 let Application = require('./src/browser/api/application.js').Application;
 let System = require('./src/browser/api/system.js').System;
-import { Window } from './src/browser/api/window';
+import { create } from './src/browser/api/window';
```
MichaelMCoates, line 24: Nitpick: Considering this file is as huge as it is, I think it's more descriptive to have the `Window.create` call vs just a plain `create` call. Can be hard to tell exactly what's being created.

pbaize: Trying not to import the whole file here, it helps with circular dependencies on the typescript conversion

MichaelMCoates: Ah, good call. Maybe hit it with an `import { create as createWindow} from './src/browser/api/window';` ?

MichaelMCoates (approved):
LGTM

PR #991, opened 2019-10-11
RUN-5666 - Child Windows showing with delay when launched from anchor tags

luiemilio:
#### Description of Change

Child windows are being registered in coreState after trying to show them in the windowPositioningObserver. This leads to the window not being found and then only showing once the window was loaded. I moved the observer logic after the coreState registration, which does the trick.

#### Checklist

- [X] `npm test` passes
- [X] PR release notes describe the change in a way relevant to app-developers

#### Release Notes

Notes: 

Fixed issue where windows opened through anchor tags have a delay

### DIFF
--- a/src/browser/api/window.js
+++ b/src/browser/api/window.js
@@ -433,6 +433,17 @@ Window.create = function(id, opts) {
 
     let _externalWindowEventAdapter;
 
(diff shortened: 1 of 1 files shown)


MichaelMCoates:
got a conflict.

PR #995, opened 2019-10-11
RUN-5682 Return type of `getEntityIdentity` should reflect the payload it returns

aziz512:
#### Description of Change
As per the Jira issue: 
>Currently, the `getEntityIdentity` call returns a `ProviderIdentity` type. However, the function only returns`{uuid, name, isExternal}`.

This PR adds a new type for the return value of `getEntityIdentity()`

#### Checklist
- [x] `npm test` passes


src/browser/core_state.ts
```diff
@@ -125,11 +125,11 @@ export function getEntityInfo(identity: Shapes.Identity) {
...
-export function getEntityIdentity(identity: Shapes.Identity): Shapes.ProviderIdentity|undefined {
+export function getEntityIdentity(identity: Shapes.Identity): Shapes.EntityIdentity | undefined {
     const { uuid, name, entityType, parentFrame } = identity;
     const externalConn = getExternalAppObjByUuid(uuid);
     if (externalConn) {
-        return {...externalConn, isExternal: true };
+        return { ...externalConn, isExternal: true };
     }
 
     const ofWindow = getWindowByUuidName(uuid, name);
```
pbaize, line 135: We should make this work for browserviews while we're at it

dhamberlin (dismissed):

src/shapes.ts
```diff
@@ -25,6 +25,10 @@ export interface ProviderIdentity extends Identity {
     runtimeUuid?: string;
 }
 
+export interface EntityIdentity extends Identity {
```
licui3936: It just adds an extra property 'isExternal'.  Can we make this property as an optional one in Identity so we don't need this new type?

pbaize: It's probably better to only have this property defined where it might be set rather than changing identity everywhere. Makes the type more meaningful.

licui3936: Agree to make the type more meaningful.

aziz512: Fixed by inline extending of Identity instead of creating a new type

src/browser/core_state.ts
aziz512: @pbaize Done!

